PR #81, opened 2020-02-28
Ahoyapps 338 community variant

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-338](https://issues.corp.twilio.com/browse/AHOYAPPS-338)

### Description

This PR adds passcode auth to the video app. It can be enabled at build time with an environment variable:
`REACT_APP_SET_AUTH=passcode`

Firebase auth is still available:

`REACT_APP_SET_AUTH=firebase`

This passcode auth is meant to be used with a Twilio Function that can be deployed using a Twilio CLI Plugin currently in development. 

image

image

These screens ^ are only visible if the passcode is not present in the URL. 

http://my-video-url.com/?passcode=1234567890

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/MenuBar/MenuBar.tsx
+++ b/src/components/MenuBar/MenuBar.tsx
@@ -66,7 +66,10 @@ export default function MenuBar() {
 
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
-    window.history.replaceState(null, '', window.encodeURI(`/room/${roomName}`));
+    // If this app is deployed as a twilio function, don't change the URL beacuse routing isn't supported.
+    if (!window.location.origin.includes('twil.io')) {
+      window.history.replaceState(null, '', window.encodeURI(`/room/${roomName}`));
+    }
     getToken(name, roomName).then(token => connect(token));
   };
 
--- a/src/state/index.tsx
+++ b/src/state/index.tsx
@@ -16,6 +16,15 @@ export interface StateContextType {
 
 export const StateContext = createContext<StateContextType>(null!);
 
+/*
+  The 'react-hooks/rules-of-hooks' linting rules prevent React Hooks fron being called
+  inside of if() statements. This is because hooks must always be called in the same order
+  every time a component is rendered. The 'react-hooks/rules-of-hooks' rule is disabled below
+  because the "if (process.env.REACT_APP_SET_AUTH === 'firebase')" statements are evaluated
+  at build time (not runtime). If the statement evaluates to false, then the code is not
+  included in the bundle that is produced (due to tree-shaking). Thus, in this instance, it
+  is ok to call hooks inside if() statements.
+*/
 export default function AppStateProvider(props: React.PropsWithChildren<{}>) {
   const [error, setError] = useState<TwilioError | null>(null);
   let contextValue = {
--- a/src/state/index.tsx
+++ b/src/state/index.tsx
@@ -8,7 +8,7 @@ export interface StateContextType {
   error: TwilioError | null;
   setError(error: TwilioError | null): void;
   getToken(name: string, room: string, passcode?: string): Promise<string>;
-  user?: User | null | { displayName: undefined; photoURL: undefined, passcode?:string };
+  user?: User | null | { displayName: undefined; photoURL: undefined; passcode?: string };
   signIn?(passcode?: string): Promise<void>;
   signOut?(): Promise<void>;
   isAuthReady?: boolean;
@@ -23,12 +23,12 @@ export default function AppStateProvider(props: React.PropsWithChildren<{}>) {
     setError,
   } as StateContextType;
 
-  if (process.env.REACT_APP_USE_FIREBASE_AUTH === 'true') {
+  if (process.env.REACT_APP_SET_AUTH === 'firebase') {
     contextValue = {
       ...contextValue,
       ...useFirebaseAuth(), // eslint-disable-line react-hooks/rules-of-hooks
     };
-  } else if (process.env.REACT_APP_USE_PASSCODE_AUTH === 'true') {
+  } else if (process.env.REACT_APP_SET_AUTH === 'passcode') {
     contextValue = {
       ...contextValue,
       ...usePasscodeAuth(), // eslint-disable-line react-hooks/rules-of-hooks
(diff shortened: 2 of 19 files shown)


src/state/index.tsx
```diff
@@ -1,52 +1,67 @@
...
   setError(error: TwilioError | null): void;
-  getToken(name: string, room: string): Promise<string>;
-  user: User | null;
-  signIn(): Promise<void>;
-  signOut(): Promise<void>;
-  isAuthReady: boolean;
+  getToken(name: string, room: string, passcode?: string): Promise<string>;
+  user?: User | null | { displayName: undefined; photoURL: undefined; passcode?: string };
+  signIn?(passcode?: string): Promise<void>;
+  signOut?(): Promise<void>;
```
charliesantos: It seems like we're going to be doing optional chaining whenever we use signIn and signOut anywhere.  Is there a way to handle this inside the function so we don't have to do `signOut?()` or `signIn?()` everywhere?

src/components/MenuBar/MenuBar.tsx
```diff
@@ -66,7 +66,10 @@ export default function MenuBar() {
 
   const handleSubmit = (event: FormEvent<HTMLFormElement>) => {
     event.preventDefault();
-    window.history.replaceState(null, '', window.encodeURI(`/room/${roomName}`));
+    // If this app is deployed as a twilio function, don't change the URL beacuse routing isn't supported.
+    if (!window.location.origin.includes('twil.io')) {
```
charliesantos: This is ok for now. But not fool proof since domain can change. Another way (potentially better) is to detect if you're running in a twilio function environment then set a variable before returning the page. Kinda how server-side rendering engine works. I'm ok to punt it if you think that takes time to implement.

src/state/index.tsx
```diff
@@ -1,52 +1,67 @@
...
 
+/*
+  The 'react-hooks/rules-of-hooks' linting rules prevent React Hooks fron being called
+  inside of if() statements. This is because hooks must always be called in the same order
+  every time a component is rendered. The 'react-hooks/rules-of-hooks' rule is disabled below
+  because the "if (process.env.REACT_APP_SET_AUTH === 'firebase')" statements are evaluated
+  at build time (not runtime). If the statement evaluates to false, then the code is not
+  included in the bundle that is produced (due to tree-shaking). Thus, in this instance, it
+  is ok to call hooks inside if() statements.
+*/
```
charliesantos: Nice way to explain this 👍

src/state/usePasscodeAuth/usePasscodeAuth.ts
```diff
@@ -0,0 +1,95 @@
...
+
+export function getErrorMessage(message: string) {
+  switch (message) {
+    case 'passcode incorrect':
+      return 'Passcode is incorrect';
+    case 'passcode expired':
+      return 'Passcode has expired';
+    default:
+      return message;
+  }
```
charliesantos: This function seems weird. Should we define error codes instead? This comes from the twilio function endpoint right?

src/state/usePasscodeAuth/usePasscodeAuth.ts
```diff
@@ -0,0 +1,95 @@
...
+    method: 'POST',
+    headers: {
+      'content-type': 'application/json',
+    },
+    body: JSON.stringify({ user_identity: name, room_name: room, passcode }),
+  });
+}
+
+export function verifyPasscode(passcode: string) {
+  return fetchToken('verification name', 'verification room', passcode).then(async res => {
```
charliesantos: Are `'verification name', 'verification room'` temporary identity and room values? If so, what do you think about renaming to `temp-identity` and `temp-room`?

timmydoza: Sounds good. :+1: They are temporary.

src/state/usePasscodeAuth/usePasscodeAuth.ts
timmydoza: The `message` property on the `error` object is basically serving as an error code. We decided on this format here: https://github.com/twilio-labs/plugin-rtc/pull/3#discussion_r387772234

src/state/index.tsx
timmydoza: I think `signIn?.()` is appropriate, because if `process.env.REACT_APP_SET_AUTH` is undefined (in the case that the user is using the local token server), then there actually are no signIn or signOut functions.

However, the UI elements that call the signIn and signOut functions are not even rendered when `REACT_APP_SET_AUTH` is undefined. So this may be a case where we can use `signIn!()` as a way to tell TypeScript "trust me, signIn will definitely be a function". 

I'm still leaning towards optional chaining though. I'm not sure it makes a big difference either way - it just feels a little safer to use. Without it I'd be worried about `signIn is not a function` errors. It also feels more descriptive. Those functions _can_ be undefined. 

What do you think?

src/components/MenuBar/MenuBar.tsx
timmydoza: I'll think about solutions for this, but I don't think we can use a server-side rendering solution. There's no server under our control that is serving the static web assets.  It's kinda like viewing a site that is hosted from an S3 bucket.

A build-time environment variable could be an option, but that has the risk that maybe the user will not set it correctly. I'll try to think of a better solution here, but for now I'm leaning towards punting it 🏈 .

src/state/usePasscodeAuth/usePasscodeAuth.ts
charliesantos: Ok, I left a comment in that PR.

src/state/index.tsx
charliesantos: I also think optional chaining sounds better over `signIn!()`. I was just concerned this might get used everywhere and see a bunch of optional chaining. Though not always bad, but usually a code smell if there are too many. I'm ok to keep it as is with the current use case.

src/components/MenuBar/MenuBar.tsx
charliesantos: Ok sounds good.

PR #88, opened 2020-03-12
AHOYAPPS-435: Readme consistency

timrozum:
https://issues.corp.twilio.com/browse/AHOYAPPS-435

1. The really obvious changes to make are grouped in the first commit.
1. Not so obvious changes were done in separate commits so they are itemized and can be reverted if necessary. See commit message for description of what changed and why.
1. I will make comments for other changes I didn't think I should make myself without discussion.
1. I used `shield` style for CircleCI badge because that is what Ryan used for iOS readme and it seems more popular on GitHub.
1. I didn't add more screenshots besides the main app screenshot because nothing that would add value came to mind and they are difficult to maintain.

### DIFF
--- a/README.md
+++ b/README.md
@@ -155,7 +155,7 @@ This application can be configured to authenticate users before they use the app
 
 See [.env.example](.env.example) for an explanation of the environment variables that must be set to enable Google authentication.
 
-## Other Platforms
+## Related
 
 - [Twilio Video Android App](https://github.com/twilio/twilio-video-app-android)
 - [Twilio Video iOS App](https://github.com/twilio/twilio-video-app-ios)
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ This application demonstrates a multi-party video application built with [twilio
 - [x] Screen sharing
 - [x] [Dominant speaker](https://www.twilio.com/docs/video/detecting-dominant-speaker) indicator
 - [x] [Network quality](https://www.twilio.com/docs/video/using-network-quality-api) indicator
-- [x] [Bandwidth Profile API](https://www.twilio.com/docs/video/tutorials/using-bandwidth-profile-api)
 
 ## Requirements
 
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# Twilio React Video App
+# Twilio Video React App
 
-[![CircleCI](https://circleci.com/gh/twilio/twilio-video-app-react.svg?style=svg&circle-token=9d6b1e89d148181aaa6874c29849c730b8ca406d)](https://circleci.com/gh/twilio/twilio-video-app-react)
+[![CircleCI](https://circleci.com/gh/twilio/twilio-video-app-react.svg?style=shield&circle-token=9d6b1e89d148181aaa6874c29849c730b8ca406d)](https://circleci.com/gh/twilio/twilio-video-app-react)
 
 This application demonstrates a multi-party video application built with [twilio-video.js](https://github.com/twilio/twilio-video.js) and [Create React App](https://github.com/facebook/create-react-app).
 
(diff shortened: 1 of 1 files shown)


timrozum:
@timmydoza take a look at the `For Twilions` section in other readmes. It seems like we should do the same for react and move content for Twilio employees to a separate file. I think the details about E2E testing and Firebase info can be moved out of this file. In mobile readmes we just say that UI (aka E2E) tests can only be run by Twilio employees because they require credentials. I assume that is the case with React but could be wrong.

timrozum:
I'm not suggesting we change this but for sake of completeness wanted to point out that this read me has application architecture and configuration top level sections which the other readmes do not. Seems like this is ok.

timrozum:
@timmydoza sounds like some information about passcode expiration should be added also. Other readmes have this but I wasn't sure how different it would be for this one.

README.md
```diff
@@ -1,20 +1,19 @@
-# Twilio React Video App
+# Twilio Video React App
 
-[![CircleCI](https://circleci.com/gh/twilio/twilio-video-app-react.svg?style=svg&circle-token=9d6b1e89d148181aaa6874c29849c730b8ca406d)](https://circleci.com/gh/twilio/twilio-video-app-react)
+[![CircleCI](https://circleci.com/gh/twilio/twilio-video-app-react.svg?style=shield&circle-token=9d6b1e89d148181aaa6874c29849c730b8ca406d)](https://circleci.com/gh/twilio/twilio-video-app-react)
```
timmydoza: While we're at it, the token isn't needed anymore because its public. Can you change it to:
```
[![CircleCI](https://circleci.com/gh/twilio/twilio-video-app-react.svg?style=shield)](https://circleci.com/gh/twilio/twilio-video-app-react)
```

README.md
```diff
@@ -1,20 +1,19 @@
...
-- Screen sharing
-- [Dominant speaker](https://www.twilio.com/docs/video/detecting-dominant-speaker) detection
-- [Network quality](https://www.twilio.com/docs/video/using-network-quality-api) indicator
-- [Bandwidth Profile API](https://www.twilio.com/docs/video/tutorials/using-bandwidth-profile-api)
+- [x] Video conferencing with real-time video and audio
+- [x] Enable/disable camera
+- [x] Mute/unmute mic
+- [x] Screen sharing
+- [x] [Dominant speaker](https://www.twilio.com/docs/video/detecting-dominant-speaker) indicator
+- [x] [Network quality](https://www.twilio.com/docs/video/using-network-quality-api) indicator
```
timmydoza: I'm ok to add the boxes but let's keep Bandwidth Profile in there.

README.md
```diff
@@ -156,6 +155,11 @@ This application can be configured to authenticate users before they use the app
 
 See [.env.example](.env.example) for an explanation of the environment variables that must be set to enable Google authentication.
 
+## Related
```
timmydoza: Good idea!

README.md
timrozum: Really? The GitHub project is public but CircleCI is private right?

README.md
timrozum: Done

README.md
timrozum: After testing I think you are correct! I can view status on circle but can't modify anything. Very nice.

timrozum: Done

README.md
timrozum: Thanks I just came up with the section name. I think it was Aaron's idea to cross link.

timmydoza:
>@timmydoza take a look at the For Twilions section in other readmes. It seems like we should do the same for React and move content for Twilio employees to a separate file. I think the details about E2E testing and Firebase info can be moved out of this file. In mobile readmes we just say that UI (aka E2E) tests can only be run by Twilio employees because they require credentials. I assume that is the case with React but could be wrong.

For the React app I don't think that it needs a 'For Twilions' section. There isn't any difference for Twilion users vs. regular users.  Regular users can even run the E2E tests because you need the same credentials that are used to run the app.  Regular users can even use Firebase if they wanted to.

timrozum:
Regarding the Twilions section that makes sense. Good enough for me! Sure with I could resolve that comment. 🙄

timmydoza:
>@timmydoza sounds like some information about passcode expiration should be added also. Other readmes have this but I wasn't sure how different it would be for this one.

Good idea. I added a short statement about passcode expiration.

README.md
```diff
@@ -156,6 +158,11 @@ This application can be configured to authenticate users before they use the app
...
+
+- [Twilio Video Android App](https://github.com/twilio/twilio-video-app-android)
+- [Twilio Video iOS App](https://github.com/twilio/twilio-video-app-ios)
```
aaalaniz:
```suggestion
- [Twilio Video iOS App](https://github.com/twilio/twilio-video-app-ios)
- [Twilio CLI RTC Plugin](https://github.com/twilio-labs/plugin-rtc)
```

PR #100, opened 2020-03-20
Ahoyapps 458 bandwidth profile improvements

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-458](https://issues.corp.twilio.com/browse/AHOYAPPS-458)

### Description

This PR makes the following changes:
- Updates twilio-video to 2.3.0
- Sets `dominantSpeakerPriority` to `standard`.
- Changes `standard` resolution to `height: 720, width: 1280`

As far as track priority goes:
- All video tracks are published with 'low' priority.
- All screen share tracks are published with 'high' priority.
- `track.setPriority('high')` is only called when the user pins a participant's thumbnail. `track.setPriority(null)` is called when the participant is unpinned. 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/MainParticipant/MainParticipant.tsx
+++ b/src/components/MainParticipant/MainParticipant.tsx
@@ -1,17 +1,14 @@
 import MainParticipantInfo from '../MainParticipantInfo/MainParticipantInfo';
 import ParticipantTracks from '../ParticipantTracks/ParticipantTracks';
 import React from 'react';
-import useDominantSpeaker from '../../hooks/useDominantSpeaker/useDominantSpeaker';
 import useMainSpeaker from '../../hooks/useMainSpeaker/useMainSpeaker';
-import useScreenShareParticipant from '../../hooks/useScreenShareParticipant/useScreenShareParticipant';
+import useSelectedParticipant from '../VideoProvider/useSelectedParticipant/useSelectedParticipant';
 
 export default function MainParticipant() {
   const mainParticipant = useMainSpeaker();
-  const dominantSpeaker = useDominantSpeaker();
-  const screenShareParticipant = useScreenShareParticipant();
+  const [selectedParticipant] = useSelectedParticipant();
 
-  const videoPriority =
-    mainParticipant !== dominantSpeaker && mainParticipant !== screenShareParticipant ? 'high' : null;
+  const videoPriority = mainParticipant === selectedParticipant ? 'high' : null;
 
   return (
     /* audio is disabled for this participant component because this participant's audio 
--- a/src/components/MainParticipant/MainParticipant.tsx
+++ b/src/components/MainParticipant/MainParticipant.tsx
@@ -2,14 +2,19 @@ import MainParticipantInfo from '../MainParticipantInfo/MainParticipantInfo';
 import ParticipantTracks from '../ParticipantTracks/ParticipantTracks';
 import React from 'react';
 import useMainSpeaker from '../../hooks/useMainSpeaker/useMainSpeaker';
+import useSelectedParticipant from '../VideoProvider/useSelectedParticipant/useSelectedParticipant';
 
 export default function MainParticipant() {
   const mainParticipant = useMainSpeaker();
+  const [selectedParticipant] = useSelectedParticipant();
+
+  const videoPriority = mainParticipant === selectedParticipant ? 'high' : null;
+
   return (
     /* audio is disabled for this participant component because this participant's audio 
        is already being rendered in the <ParticipantStrip /> component.  */
     <MainParticipantInfo participant={mainParticipant}>
-      <ParticipantTracks participant={mainParticipant} disableAudio enableScreenShare videoPriority="high" />
+      <ParticipantTracks participant={mainParticipant} disableAudio enableScreenShare videoPriority={videoPriority} />
     </MainParticipantInfo>
   );
 }
--- a/src/components/MainParticipant/MainParticipant.tsx
+++ b/src/components/MainParticipant/MainParticipant.tsx
@@ -1,14 +1,17 @@
 import MainParticipantInfo from '../MainParticipantInfo/MainParticipantInfo';
 import ParticipantTracks from '../ParticipantTracks/ParticipantTracks';
 import React from 'react';
+import useDominantSpeaker from '../../hooks/useDominantSpeaker/useDominantSpeaker';
 import useMainSpeaker from '../../hooks/useMainSpeaker/useMainSpeaker';
-import useSelectedParticipant from '../VideoProvider/useSelectedParticipant/useSelectedParticipant';
+import useScreenShareParticipant from '../../hooks/useScreenShareParticipant/useScreenShareParticipant';
 
 export default function MainParticipant() {
   const mainParticipant = useMainSpeaker();
-  const [selectedParticipant] = useSelectedParticipant();
+  const dominantSpeaker = useDominantSpeaker();
+  const screenShareParticipant = useScreenShareParticipant();
 
-  const videoPriority = mainParticipant === selectedParticipant ? 'high' : null;
+  const videoPriority =
+    mainParticipant !== dominantSpeaker && mainParticipant !== screenShareParticipant ? 'high' : null;
 
   return (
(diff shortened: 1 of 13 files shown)


src/components/MainParticipant/MainParticipant.tsx
```diff
@@ -2,14 +2,19 @@ import MainParticipantInfo from '../MainParticipantInfo/MainParticipantInfo';
...
+  const [selectedParticipant] = useSelectedParticipant();
+
+  const videoPriority = mainParticipant === selectedParticipant ? 'high' : null;
+
   return (
     /* audio is disabled for this participant component because this participant's audio 
        is already being rendered in the <ParticipantStrip /> component.  */
     <MainParticipantInfo participant={mainParticipant}>
-      <ParticipantTracks participant={mainParticipant} disableAudio enableScreenShare videoPriority="high" />
+      <ParticipantTracks participant={mainParticipant} disableAudio enableScreenShare videoPriority={videoPriority} />
```
lwilio: Can I understand the rationale behind this line? Does this mean that the dominant speaker is assigned 'high' priority once it's rendered in the focus area (i.e. the main area of the screen) after the dominant-speaker event is received? If that's the case, we should not assign high-priority to the dominant speaker. The dominant-speaker priority should be only managed by VMS.

src/components/ParticipantTracks/ParticipantTracks.tsx
```diff
@@ -8,7 +8,7 @@ interface ParticipantTracksProps {
   participant: LocalParticipant | RemoteParticipant;
   disableAudio?: boolean;
   enableScreenShare?: boolean;
-  videoPriority?: Track.Priority;
+  videoPriority?: Track.Priority | null;
```
lwilio: Can you also explain the rationale behind this? Does this mean that when a video track is no longer at the focus it gets assigned 'null' as its subscribe priority?

timmydoza: When `videoPriority` is `null`, `track.setPriority` never gets called. 

I only added this change because it seems better to set `videoPriority` to `null` instead of `undefined`.

src/components/MainParticipant/MainParticipant.tsx
timmydoza: Here, `videoPriority` is set to 'high' only when the user pins a participants thumbnail. When the dominant speaker is to be rendered in the main video area, `videoPriority` will be `null` and `track.setPriority` will never be called. 

I suppose there could be a case when the user pins the dominant speaker.  In this case `track.setPriority('high')` _would_ be called on the dominant speaker's video track.  I feel like this would be an uncommon occurrence, but I can add some logic to avoid this if you think it would be worthwhile.

src/components/ParticipantTracks/ParticipantTracks.tsx
manjeshbhargav: @timmydoza We should find a way to differentiate between a VideoTrack that was added to the main window due to a dominant speaker change event vs the VideoTrack that was added to the main window due to the user pinning a Participant. In the former case, we should not be calling `setPriority()` at all. In the latter case, we should call `setPriority('high')` when pinned and `setPriority(null)` when unpinned.

timmydoza: @Sounds good. I've updated the logic so that the priority is set to high only when the main participant is not the dominant speaker or sharing their screen (because screen share tracks are already published with high priority).  `setPriority(null)` will be only be called when pinned participants are unpinned.

src/components/MainParticipant/MainParticipant.tsx
lwilio: @timmydoza thanks for the explanation. You can keep it as it is, no need to change anything. Just FYI, dominantSpeakerPriority only upgrades the priority, never downgrades it. In other words, if the user pins a video and that video becomes dominant speaker, the priority will stay as 'high' even if dominantSpeakerPriority='standard'. Hence, IMO your logic works right.

src/components/ParticipantTracks/ParticipantTracks.tsx
lwilio: @timmydoza one question that comes to may mind is what happens with the screenshare when the end-user pins another video. If I'm not wrong, the screenshare is published with priority 'high', which is OK because buy default it goes to the focus (i.e. the main area of the screen). However, if an end-user pins a different video, the screenshare should be downgraded to 'low' given that it's not being rendered. If we keep it as 'high', then it will steal relevant BW (and quality) to the pinned video while it's pinned. Is this how things work in this patch?

timmydoza: Thats correct. We used to have screen share tracks published with low priority, and then set their priority to high while they are displayed in the main video area. 

We then set the screen share tracks to always have high priority to address an issue found in this ticket: https://issues.corp.twilio.com/browse/VMS-2462. Users were reporting that screenshare tracks were getting switched off. 

It seems that the reported issue was due to a `setPriority` bug in the JS SDK which was fixed in version 2.3.0.  

I'm thinking we go back to how we used to handle screen share priority.  We publish screen share tracks with low priority, and then set the priority to high when its displayed in the main video area. What do you think @lwilio?

src/components/MainParticipant/MainParticipant.tsx
timmydoza: I see.  I did update the logic so that `setPriority` is never called on a dominant speaker track.  Do you think I should revert that so a pinned dominant speaker will have high priority instead of standard?

src/components/ParticipantTracks/ParticipantTracks.tsx
lwilio: @timmydoza If the bug is fixed IMO we should do it the right way and set the screenshare priority to low when it's not visible. Otherwise we'll be spoiling valuable BW for nothing. Of course, please double check first that the bug is really fixed just in case.

src/components/MainParticipant/MainParticipant.tsx
timmydoza: I agree that it would be simpler to always have pinned video be high priority.  I'll update the code accordingly.

src/components/ParticipantTracks/ParticipantTracks.tsx
timmydoza: @lwilio It appears to be fixed: https://github.com/twilio/twilio-video.js/blob/master/CHANGELOG.md#bug-fixes

Should I 
1) publish screen share tracks with 'low' priority and set them to 'high' when they are rendered in the main video area?
or
2) publish screen share tracks with 'high' priority, and set them to 'low' when they are not rendered in the main video area?

I'd very much prefer to go with #1 as it is much simpler to code, but I'm curious about what you think.

lwilio: @timmydoza IMO there will be currently no difference between 1 and 2 from the QoE perspective. However, I'd prefer 2 because in the future track priorities would also be considered at the publisher side (currently are only considered at the subscriber side). Also, the default behavior is to show the screenshare in the focus, hence, IMO it's clear if it's set to 'high' "by default" and we only set it to low when something "special" happens (i.e. the end-user pins another video). However, again, up to you to choose.

timmydoza: Thanks for your input @lwilio. I've thought about this one a lot, and for now I'd like to go with option #1. It's easier and faster to implement. I'm in favor of trying option 2 in the near future.

timmydoza:
@lwilio @manjeshbhargav - I've updated the PR:

- Screenshare tracks are published with 'low' priority. This should work properly again now that the `setPriority` issue is fixed in SDK version 2.3.0.
- Sets `track.setPriority('high')` when a participant is pinned, or when a participant is sharing their screen. It will call `track.setPrioirty(null)` when the video track of a screen share participant or pinned participant is unmounted. `setPriority` will never be called in response to dominant speaker events. 

Please let me know what you think.  Thanks!!

PR #119, opened 2020-03-30
Typescript fix

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-452](https://issues.corp.twilio.com/browse/AHOYAPPS-452)

### Description

Some users were seeing typescript errors while building the app (which would stop the build). The errors would occur when version 3.8.3 of Typescript is used, but not when version 3.7.5 of Typescript is used. 

These users installed dependencies with `yarn` instead of `npm`. Yarn would ignore `package-lock.json` and it would install the newest major version of Typescript (3.8.3) as described by `package.json` instead of the version of TS that is specified in `package-lock.json` (3.7.5).

This PR updates TS to version 3.8.3 and fixes the errors. Now the app can successfully be built even when users install dependencies with `yarn`. 

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,8 @@ You must have the following installed:
 
 Run `npm install` to install all dependencies from NPM.
 
+If you want to use `yarn` to install dependencies, first run the [yarn import](https://classic.yarnpkg.com/en/docs/cli/import/) command. This will ensure that yarn installs the package versions that are specified in `package-lock.json`.
+
 ## Install Twilio CLI
 
 The app is deployed to Twilio using the Twilio CLI. Install twilio-cli with:
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,37 +13,51 @@
       }
     },
     "@babel/compat-data": {
-      "version": "7.8.5",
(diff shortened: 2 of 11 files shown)


timmydoza:
> The changes look good. Should we also update the documentation/README around [here](https://github.com/twilio/twilio-video-app-react#install-dependencies) and mention something related to this? Something like if you use a package manager such as yarn, it might not pick up versions specified in package-lock.json which could cause issues.
> 
> I can see this happening in the future if we don't warn the customers.

Sure! What do you think about this:

If you want to use `yarn` to install dependencies, first run the [yarn import](https://classic.yarnpkg.com/en/docs/cli/import/) command. This will ensure that yarn installs the package versions that are specified in `package-lock.json`.

charliesantos:
> Sure! What do you think about this:
>
> If you want to use yarn to install dependencies, first run the yarn import command. This will ensure that yarn installs the package versions that are specified in package-lock.json.

Sounds good to me 👍

PR #129, opened 2020-04-03
Audio track render

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-514](https://issues.corp.twilio.com/browse/AHOYAPPS-514)

### Description

This PR adds:
- Passing `?customIdentity=true` will allow users to edit their Identity when using firebase auth (very useful for testing)
- Exposes `Video` object as `window.TwilioVideo`
- Adds an Audio Level Indicator component to the menubar, which lets users visually verify that their microphone is working before they join a room
- Adds Audio Level Indicators to each participant. It shows their audio level, which is useful to see who is speaking, and it's also useful for finding audio issues. The Audio indicator also indicates when participant's audio is muted. 
- Improves mobile layout when joining a room so that mobile users can easily see the Audio Indicator

image

image

image

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
+++ b/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
@@ -3,6 +3,9 @@ import { AudioTrack } from 'twilio-video';
 import MicOff from '@material-ui/icons/MicOff';
 import useIsTrackEnabled from '../../hooks/useIsTrackEnabled/useIsTrackEnabled';
 
+// @ts-ignore
+const AudioContext = window.AudioContext || window.webkitAudioContext;
+
 export default function AudioLevelIndicator({
   size,
   audioTrack,
@@ -61,7 +64,7 @@ export default function AudioLevelIndicator({
     <svg focusable="false" viewBox="0 0 24 24" aria-hidden="true" height={`${SIZE}px`} width={`${SIZE}px`}>
       <defs>
         <clipPath id="audio-level-clip">
-          <rect ref={ref} x="0" y="4" width="24" height="24" />
+          <rect ref={ref} x="0" y="21" width="24" height="24" />
         </clipPath>
       </defs>
       <path
--- a/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
+++ b/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
@@ -10,6 +10,21 @@ const getUniqueClipId = () => clipId++;
 const AudioContext = window.AudioContext || window.webkitAudioContext;
 let audioContext: AudioContext;
 
+export function initializeAnalyser(audioTrack: AudioTrack) {
+  audioContext = audioContext || new AudioContext();
+  const stream = new MediaStream();
+  const track = audioTrack.mediaStreamTrack.clone();
+  stream.addTrack(track);
+  const audioSource = audioContext.createMediaStreamSource(stream);
+
+  const analyser = audioContext.createAnalyser();
+  analyser.smoothingTimeConstant = 0.5;
+  analyser.fftSize = 1024;
+
+  audioSource.connect(analyser);
+  return analyser;
+}
+
 function AudioLevelIndicator({
   size,
   audioTrack,
@@ -28,16 +43,16 @@ function AudioLevelIndicator({
     let animationFrameID: number;
 
     if (audioTrack && isTrackEnabled && SVGClipElement) {
-      audioContext = audioContext || new AudioContext();
-      const stream = new MediaStream();
-      stream.addTrack(audioTrack.mediaStreamTrack.clone());
-      const audioSource = audioContext.createMediaStreamSource(stream);
+      let analyser = initializeAnalyser(audioTrack);
 
-      const analyser = audioContext.createAnalyser();
-      analyser.smoothingTimeConstant = 0.5;
-      analyser.fftSize = 1024;
+      const reinitializeAnalyser = () => {
+        analyser = initializeAnalyser(audioTrack);
+      };
 
-      audioSource.connect(analyser);
+      // Here we reinitialize the AnalyserNode on focus to avoid an issue in Safari
+      // where the analysers stop functioning when the user switches to a new tab
+      // and switches back to the app.
+      window.addEventListener('focus', reinitializeAnalyser);
 
       function render() {
         animationFrameID = window.requestAnimationFrame(() => {
@@ -62,6 +77,7 @@ function AudioLevelIndicator({
       return () => {
         SVGClipElement.setAttribute('y', '21');
         window.cancelAnimationFrame(animationFrameID);
+        window.removeEventListener('focus', reinitializeAnalyser);
       };
     }
   }, [audioTrack, isTrackEnabled]);
--- a/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
+++ b/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
@@ -1,0 +1,70 @@
+import React, { useEffect, useRef } from 'react';
+import { LocalAudioTrack } from 'twilio-video';
+import useVideoContext from '../../hooks/useVideoContext/useVideoContext';
+
+export default function({ size }: { size?: number }) {
+  const SIZE = size || 24;
+  const ref = useRef<SVGRectElement>(null);
+  const { localTracks } = useVideoContext();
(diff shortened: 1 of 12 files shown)


src/components/MenuBar/MenuBar.tsx
```diff
@@ -78,7 +89,7 @@ export default function MenuBar() {
       <Toolbar>
         {roomState === 'disconnected' ? (
           <form className={classes.form} onSubmit={handleSubmit}>
-            {!user?.displayName ? (
+            {window.location.search.includes('customIdentity=true') || !user?.displayName ? (
```
charliesantos, line 92: How many query parameters do we parse on the URL now? I'm wondering if it might be a good time to create a utility function that returns the value of a query param. Example: `const customIdentity = getQueryParam('customIdentity')`. WDYT?

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
```diff
@@ -0,0 +1,99 @@
...
+      const audioSource = audioContext.createMediaStreamSource(stream);
+
+      const analyser = audioContext.createAnalyser();
+      analyser.smoothingTimeConstant = 0.5;
+      analyser.fftSize = 1024;
+
+      audioSource.connect(analyser);
+
+      function render() {
+        animationFrameID = window.requestAnimationFrame(() => {
```
charliesantos, line 43: requestAnimationFrame doesn't work if there are no repaints on the browser. For example, out of focus tabs will not get this. We solved this in voice by using a setTimeout

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
```diff
@@ -0,0 +1,99 @@
...
+          const array = new Uint8Array(analyser.frequencyBinCount);
+          analyser.getByteFrequencyData(array);
+          let values = 0;
+
+          const length = array.length;
+          for (let i = 0; i < length; i++) {
+            values += array[i];
+          }
+
+          const volume = Math.min(21, Math.max(0, Math.log10(values / length / 3) * 14));
```
charliesantos, line 53: It feels like getting volume levels should be part of the SDK. Is this API not supported yet?

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
charliesantos: We should also throttle this because this eats up a lot of CPU which makes the app slow. Very noticeable on low resource machines or phones.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
```diff
@@ -0,0 +1,99 @@
...
+    const SVGClipElement = ref.current;
+    let animationFrameID: number;
+
+    if (audioTrack && isTrackEnabled && SVGClipElement) {
+      audioContext = audioContext || new AudioContext();
+      const stream = new MediaStream();
+      stream.addTrack(audioTrack.mediaStreamTrack.clone());
+      const audioSource = audioContext.createMediaStreamSource(stream);
+
+      const analyser = audioContext.createAnalyser();
```
charliesantos, line 36: Have we tested this on other browsers? I know there are audio level issues on safari, especially if the browser goes from focus -> unfocus -> focus. When this happens, you will get zero levels.

manjeshbhargav: @charliesantos This is the whole point of the indicator. When this happens, the zero levels should result in the app displaying a warning asking the customer to re-join the room.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
manjeshbhargav: @charliesantos Actually, out of focus tabs need not repaint the audio level, so this is fine. Generally `setTimeout()` is not recommended for UI updates. Even `setTimeout` is throttled when a tab is backgrounded.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
manjeshbhargav: @charliesantos Currently it is not. We are thinking of adding it though.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
charliesantos: @manjeshbhargav I understand the goal (I think). However, the issue I'm seeing with this approach is, we are using volume levels to determine whether the customer should re-join the room or not. Getting volume levels in this manner proved to have issues on Safari and not reliable. Here's an example.

I use safari, join a room. Everything is fine. My microphone works and everyone can hear me. Audio level indicator looks good also. Then, zero level happens (you can trigger by going out of focus on safari), but my mic still works and everyone can still hear me. Would we still show the warning and ask the customer to re-join the room?

Also, I have not tested the react app with this on safari. But I have tested on VoiceJS SDK because we currently have the same implementation.

Ultimately, the point I'm trying to make is, we should not rely on this across browser or if we do, since this might be the only option we have, we should be able to detect when these zero level issue happens and handle it instead of always showing the warning.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
charliesantos: Ok. This is probably fine since this is for UI. In our case as a SDK, customers wants to detect volume levels even when on background - out of focus, or when running as extension.

timmydoza: I'm ok with this. We're using `requestAnimationFrame` for animating, so if the animations stop when the tab is out of focus that is ok. 

Also I'm not worried about performance. This kind of animation appears to be very efficient. I created a demo [here](https://winter-luck-gargoyle.glitch.me/) that creates 300 similar animations and it runs smoothly on my Galaxy 8. 

At most we would have 50 participants, and this would only animate those that have unmuted audio, which would be a number much lower than 50 in most cases.

src/components/MenuBar/MenuBar.tsx
timmydoza: Right now we get query parameters here, and when we consume the passcode (when using passcode authentication). 

So for now I think this is fine, but when we implement a settings menu that (perhaps) consumes a query parameter for each setting (useful for testing), then I think implementing your idea would be good.

charliesantos: Ok sounds good. Let's keep this at the back of our head. We don't want to keep this kind of pattern used everywhere.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
charliesantos: I just tested on Safari and I was able to reproduce the issue where zero levels happen.
Steps:
1. Load app on safari. Observe the audio indicator. Should work when you speak.
2. Open a new tab and make sure the new tab is focused
3. Now click the original tab on step 1.
4. Try to speak, audio indicator stops working.

When this happens, I don't think we should just show a generic message asking the customer to re-join. We should either find a fix for this or detect when this happens so we can show a different message. I don't have a solution on how to detect the bug though.

timmydoza: @charliesantos I pushed up a fix for this.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
```diff
@@ -10,6 +10,21 @@ const getUniqueClipId = () => clipId++;
 const AudioContext = window.AudioContext || window.webkitAudioContext;
 let audioContext: AudioContext;
 
+export function initializeAnalyser(audioTrack: AudioTrack) {
+  audioContext = audioContext || new AudioContext();
+  const stream = new MediaStream();
+  const track = audioTrack.mediaStreamTrack.clone();
+  stream.addTrack(track);
+  const audioSource = audioContext.createMediaStreamSource(stream);
```
manjeshbhargav, line 18: @timmydoza 
```suggestion
  const stream = new MediaStream([audioTrack.mediaStreamTrack.clone()]);
  const audioSource = audioContext.createMediaStreamSource(stream);
```

PR #131, opened 2020-04-04
TS path aliases (absolute imports)

TugayYaldiz:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

I'm decided to add TS path aliases support (absolute imports) in my project (which the one I forked from this repo) and change relative imports like `../../../hooks/useVideoContext/useVideoContext` this to `@hooks/useVideoContext` this. I wanted to share with community as well!

aliases
---
Added four new development dependencies for this: 
(other changes in devDependencies was just rearranged by NPM itself)

`react-app-rewired` and `customize-cra` for extending `create-react-app` without ejecting .

`tsconfig-paths-webpack-plugin` for adding path aliases support. This adds support for Webpack but for Jest we need to add `moduleNameMapper` to `jest.config.js` manually like:
```javascript
moduleNameMapper: {
    '^@src(.*)$': '<rootDir>/src$1',
    '^@components(.*)$': '<rootDir>/src/components$1',
    '^@hooks(.*)$': '<rootDir>/src/hooks$1',
   ...
  }
``` 

`patch-package` for changing `react-scripts` package code for suppressing `compilerOptions.paths must not be set (aliased imports are not supported)`  this is needed for preventing CI run fails. The change can be found in `patches` folder. And I updated `.circleci/config.yml` to make sure patches are included in the cache, due to https://github.com/ds300/patch-package#docker-and-ci 

---

Had to add these packages because `create-react-app` doesn't support this feature yet [#5645](https://github.com/facebook/create-react-app/issues/5645). And doesn't seem to they will support it soon.

They are removing unsupported options in `tsconfig.json` like `paths` when the app starts in  whether build or dev mode [#5609](https://github.com/facebook/create-react-app/pull/5609). For preventing CRA's doing this we need to create another small tsconfig file `tsconfig.paths.json` and extends our base config with that https://github.com/facebook/create-react-app/issues/5645#issuecomment-461999138. But `react-scripts` still complains at you with warning  so `patch-package` is for that.  

When they support this feature or if you consider ejecting CRA in the future, you can just remove this development packages. Except `tsconfig-paths-webpack-plugin` you will probably still need this package, if you eject.

And I also changed folder structures a bit, with adding `index.ts` files and two folders also has typo `NewtorkQualityLevel` and `ToogleScreenShareButton`

All these changes is for better scalability and to make future refactoring, fixing or updating code easier.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


timmydoza:
Thanks for the PR @TugayYaldiz! 

I like the idea of having absolute paths. It can be confusing to look at `../../../..`. The TS alias is a neat feature, but I think there is a simpler way to get absolute imports. 

If you add `"baseUrl": "src"` to `compilerOptions` in `tscofig.json`, and add `modulePaths: ['src']` to `jest.config.js`, then you can use absolute paths in your imports. I have an example branch here: https://github.com/twilio/twilio-video-app-react/compare/absolute-import-example

I like approach because the configuration changes are minimal, and it doesn't require any additional dependencies. 

One potential downside is that there is no `@` symbol to show which paths are absolute, and this could be a little confusing for an import like `import { useAppState } from 'state';`.  That looks like `useAppstate` is being imported from an NPM module called `state`. This could be fixed by setting the `baseURL` to `.` instead of `src`, which would change the import to `import { useAppState } from 'src/state';`, which doesn't look like an NPM module.

What do you think about this approach?

TugayYaldiz:
Yes you can set only `baseURL` and always start from `src` but you need to start always from src. 
It doesn't make you feel like they are separated things like `components`, `hooks` etc. But making path aliases makes them path independent. 

I mean you don't care where the components are its just `@components` for you.  😄 And you may want to  change folder structures in future when the project grows larger, for example if you wanted to implement [Atomic Design Methodology](https://bradfrost.com/blog/post/atomic-web-design/), you can just simply move the folders into another folders and update it's path from `tsconfig`done. Simple is that 😄React docs also has good point on that

>There are many pain points associated with deep directory nesting in JavaScript projects. It becomes harder to write relative imports between them, or to update those imports when the files are moved....
https://en.reactjs.org/docs/faq-structure.html#avoid-too-much-nesting

Maybe you may say my IDE updates paths automatically when the files are moved, but this still does not change the fact that you have to update those paths.

Honestly I prefer configuring  path aliases and treating them like separated things, rather than having just one absolute base. But of course this is a decision that you will make anyway. 🙂

And I prefer `index.ts` files to be added also for scalability and getting rid of weird duplicated `..../useVideoContext/useVideoContext` things.

timmydoza:
Thanks for the info @TugayYaldiz! You make some very good points. I can see that path aliases provide some great benefits that absolute paths do not. 

However, it seems that path aliases solve problems that this app doesn't have. We have no plans of changing the directory structure of the app, so I don't think the benefit is worth the cost of adopting the proposed configuration changes and dependencies (26 new dependencies counting transitive ones).  Unfortunately, the added complexity (and time required to review a PR with 134 changed files) is too great for the benefits we would receive. 

I would be in favor of using absolute paths - especially since the cost is quite low (no dependencies to install - just need to update 2 json files). This isn't a high priority item for us though, because using 100% relative paths hasn't been a huge pain point. 

Thanks for pointing out the typos! I'll try to get those fixed soon. 😄

PR #138, opened 2020-04-08
Throttle AudioIndicator animation

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

* [AHOYAPPS-554](https://issues.corp.twilio.com/browse/AHOYAPPS-554)

### Description

It appears that the recent addition of the Audio Indicator component may have increased the CPU usage of some users.  This PR throttles the computation of audio levels from ~60fps to ~20fps. It uses the `interval` function from [D3's timer module](https://github.com/d3/d3-timer), which is like `setTimeout`, but it uses `requestAnimationFrame` under the hood. 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
```diff
@@ -52,29 +52,24 @@ function AudioLevelIndicator({
       // and switches back to the app.
       window.addEventListener('focus', reinitializeAnalyser);
 
-      function render() {
-        animationFrameID = window.requestAnimationFrame(() => {
-          const array = new Uint8Array(analyser.frequencyBinCount);
-          analyser.getByteFrequencyData(array);
-          let values = 0;
+      const timer = interval(() => {
```
charliesantos, line 55: What's the difference between this and setInterval?

timmydoza: It's like `setInterval`, but it uses `requestAnimationFrame` under the hood. This makes it good for animations, and it also guarantees that it won't run in the background.

timmydoza:
I added two more optimizations. I lowered the `fftsize` from 1024 to 512. This halves the number of samples to sum during each frame. I also moved `new Uint8Array()` outside of the animation loop so that one `Uint8Array` is created per component, instead of per frame.

src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
charliesantos: What is the 2nd parameter for? Is that 50ms? Basically, it will use requestAnimationFrame but it won't run more than once within a 50ms time period?

timmydoza: Correct. It's like `window.setTimeout(func, 50)`.

https://github.com/d3/d3-timer#interval

charliesantos: Ok cool. I think this looks good 👍

PR #152, opened 2020-04-14
Ahoyapps 495 switch camera

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-495](https://issues.corp.twilio.com/browse/AHOYAPPS-495)

### Description

This PR adds an icon that allows mobile users to toggle between their front and rear facing cameras. Users can do this before or after they join a room. 

IMG_0003

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/MenuBar/FlipCameraButton/FlipCameraButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+    // However, if facingMode is ever false again (when the user unpublishes video), we
+    // won't set 'supportsFacingMode' to false. This prevents the icon from briefly
+    // disappearing when the user switches their front/rear camera.
+    if (facingMode && supportsFacingMode === null) {
+      setSupportsFacingMode(Boolean(facingMode));
+    }
+  }, [facingMode, supportsFacingMode]);
+
+  const toggleFacingMode = useCallback(() => {
+    const localTrackPublication = localParticipant?.unpublishTrack(videoTrack!);
```
manjeshbhargav, line 28:
```suggestion
    const localTrackPublication = localParticipant?.unpublishTrack(videoTrack!);
    videoTrack.stop();
```

src/components/MenuBar/FlipCameraButton/FlipCameraButton.tsx
```diff
@@ -0,0 +1,45 @@
...
+    localTracks,
+    getLocalVideoTrack,
+  } = useVideoContext();
+  const [supportsFacingMode, setSupportsFacingMode] = useState<Boolean | null>(null);
+  const videoTrack = localTracks.find(track => track.name === 'camera');
+  const facingMode = videoTrack?.mediaStreamTrack.getSettings().facingMode;
+
+  useEffect(() => {
+    // The 'supportsFacingMode' variable determines if this component is rendered
+    // If facingMode is true, we will set supportsFacingMode to true.
```
charliesantos, line 18: Nit: To be more accurate, should this say "If facingMode is available (or something similar),"? Because facingMode is not boolean right?

src/components/VideoTrack/VideoTrack.tsx
```diff
@@ -35,7 +35,8 @@ export default function VideoTrack({ track, isLocal, priority }: VideoTrackProps
   }, [track, priority]);
 
   // The local video track is mirrored.
-  const style = isLocal ? { transform: 'rotateY(180deg)' } : {};
+  const isFrontFacing = track.mediaStreamTrack.getSettings().facingMode !== 'environment';
```
charliesantos, line 38: How about `track.mediaStreamTrack.getSettings().facingMode === 'user';`? Because if facingMode is `left` or `right`, this evaluates to true.

timmydoza: `facingMode` is a property that seems to exist only on mobile devices. So `isFrontFacing` would be `false` when using the app on my laptop, and the video would not be mirrored when it should be. 

Also, according to MDN, facingMode of `left` and `right` represents cameras that are front facing, just pointed over the user's left or right shoulder. I can't think of any devices with these cameras, but we would want `isFrontFacing` to be `true` in those cases.

src/components/MenuBar/FlipCameraButton/FlipCameraButton.tsx
timmydoza: Good call - updated :+1:

PR #163, opened 2020-04-17
Ahoyapps 561 mobile best practices

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-561](https://issues.corp.twilio.com/browse/AHOYAPPS-561)

### Description

This PR add the following features:
- When the user is on mobile, disconnects the app on the `pagehide` event
- Limits max incoming video bitrate to 2.5mbps
- Adds comments to the ConnectionOptions object so users can better understand the settings
- On mobile devices, adds a `visibilitychange` listener to unpublish local video when the app is backgrounded, and republishes it when the app is foregrounded. Note: video is not republished if it was unpublished before the app was backgrounded (in the case that the user mutes their video). 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/VideoProvider/AttachVisibilityHandler/AttachVisibilityHandler.tsx
```diff
@@ -0,0 +1,44 @@
...
+  to show that this user's video track has been turned off.
+*/
+
+export default function AttachVisibilityHandler() {
+  const { room } = useVideoContext();
+  const [isVideoEnabled, toggleVideoEnabled] = useLocalVideoToggle();
+  const shouldRepublishVideoOnForeground = useRef(false);
+
+  useEffect(() => {
+    if (isMobile) {
```
charliesantos, line 22: I try to stay away from browser detection if possible and use feature detection instead. Should we do that here?

timmydoza: That's a good point. For this code, I based it off of @manjeshbhargav's changes in the same quickstart feature: https://github.com/twilio/video-quickstart-js/blob/mobile-browser-support/quickstart/src/joinroom.js#L271

Is there something that you have in mind for feature detection?

charliesantos: In general, it depends on the feature we're using. In this case, it seems like we can remove the `isMobile` check. If the browser doesn't support `visibilitychange`, this code will not do anything, is that correct?

timmydoza: `visibilitychange` is supported on mobile and desktop browsers, but we only want this feature on mobile. We want to unpublish the user's video when a mobile browser is backgrounded, but we don't want to unpublish the user's video when the user changes tabs in a desktop browser. 

[This article](https://developer.mozilla.org/en-US/docs/Web/HTTP/Browser_detection_using_the_user_agent) recommends feature detection over user-agent sniffing, but in this case I think that user-agent sniffing should be sufficient. Although now I'm wondering if we should also check for [Firefox tablets](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent/Firefox#Mobile_and_Tablet_indicators) as well.

I'll ask in the quickstart PR.

charliesantos: > but we don't want to unpublish the user's video when the user changes tabs in a desktop browser.

Please help me understand why? Because this sounds like an improvement.

timmydoza: Users are often multi-tasking when using the desktop app. If you switch to another tab, do you want your local video to be unpublished from the room? Other users could only see you when you have the app in the active tab. There would be no way for users to see if you if you are using another tab. 

On mobile, when the app is backgrounded, the camera stops sending frames. So we unpublish the video track to 1) update the UI in other participant's apps to show that the mobile user's video is gone (this will display the correct icon), and 2) free up the camera resource on the mobile device when it is not in use. 

On desktop, the camera still functions perfectly fine when the user changes tabs, so there is no need to unpublish it.

charliesantos: Ok. I understand the use case.

timmydoza: Sorry, I probably should have referenced the recently published best practices guide that this PR is implementing: https://www.twilio.com/docs/video/build-js-video-application-recommendations-and-best-practices#application-backgrounding-in-mobile-browsers

Note: this guide will be amended to recommend that video tracks are unpublished, not disabled.

PR #203, opened 2020-05-14
Settings menu

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-640](https://issues.corp.twilio.com/browse/AHOYAPPS-640)

See [this ticket](https://issues.corp.twilio.com/browse/AHOYAPPS-638) for specific requirements

### Description

Adds a settings menu for setting some bandwidth profile options. 

image
image
Mobile:
image

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


charliesantos:
Implementation looks good. Just one question. Can you remind me again what's our criteria when we should use reducer? It's when we need app wide states correct? Any other reasons?

timmydoza:
> @timmydoza I checked out the screenshots. They look awesome. Is there anyway we can group settings under a "Bandwidth Profile" header.

Added.

image

timmydoza:
> Implementation looks good. Just one question. Can you remind me again what's our criteria when we should use reducer? It's when we need app wide states correct? Any other reasons?

Sure! In order to have app-wide state, we use Context. All of the values passed to the `StateContext.Provider` component (https://github.com/twilio/twilio-video-app-react/blob/settings-menu/src/state/index.tsx#L88) can then be accessed anywhere in the app using the `useAppState()` hook.  

`useReducer` stores state and updates components when the state changes. To me it's the same as `useState`, just with a different API. What's interesting to me - if you look in the React source code, `useState` uses `useReducer` under the hood. So the two functions have the same purpose, but this PR I chose to use `useReducer` because the implementation would be a little less complex.

PR #211, opened 2020-05-18
Hide ToggleFullscreenButton when not supported

jhujhul:
**Contributing to Twilio**

### Description

When using the application on iPhone, the ToggleFullscreenButton does not work. It's because iOS Safari on iPhone doesn't support the Full Screen API:
https://caniuse.com/#feat=fullscreen

I think it could be better to display the ToggleFullscreenButton only if the Full Screen API is supported by the browser. What do you think?

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/MenuBar/MenuBar.tsx
+++ b/src/components/MenuBar/MenuBar.tsx
@@ -1,5 +1,6 @@
 import React, { ChangeEvent, FormEvent, useState, useEffect } from 'react';
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
+import fscreen from 'fscreen';
 
 import AppBar from '@material-ui/core/AppBar';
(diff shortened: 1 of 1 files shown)


timmydoza:
Thank you so much for your contribution @jhujhul!!

Looks like you just beat me to it by a few minutes! I just pushed up a similar fix here: https://github.com/twilio/twilio-video-app-react/pull/212

Your PR would certainly fix the issue. However I'm leaning towards merging in #212 instead because it performs the `fscreen.fullscreenEnabled` check inside the `ToggleFullScreenButton` itself (which would allow us to add the button elsewhere in the app without having to re-implement the `fscreen.fullscreenEnabled` check), and it also has a unit test. 

Thanks again for your help!

Issue #233, opened 2020-06-05
Mute Unmute causing error and multiple tumbnails


asyfsaeed:
Switching between un mute and mute is causing weird behavior 

Please check attach screen shots

![Screenshot 2020-06-05 at 11 53 28 PM](https://user-images.githubusercontent.com/5968466/83912867-033de400-a788-11ea-824c-8a02fe35f8bf.png)
![Screenshot 2020-06-05 at 11 53 16 PM](https://user-images.githubusercontent.com/5968466/83912873-0638d480-a788-11ea-876e-7cd51cc556c7.png)

timmydoza:
Thanks for the issue @asyfsaeed! Sorry to hear that you are having issues. 

I am unable to reproduce the issue in the screenshot. It is difficult to know what the problem may be without additional information. 

Specifically, it would be very helpful to us if you could provide us with the following information:

- What steps need to be taken to reproduce this issue?
- How quickly are you switching between mute and unmute?
- What operating system and OS version are you using? 
- What browser and browser version are you using? 
- What version of the App are you using? (master branch, specific commit hash)
- What version of Node.js version are you using?

asyfsaeed:
Hi,
1. Joined room and start switching mute and unmute
2. Two seconds between clicking mute and unmute
3. AWS Ubuntu t2.medium
4. Chrome Version 83.0.4103.97 (Official Build) (64-bit)
5. latest master code
6. 10.13.0

asyfsaeed:
And sometime camera  video remains on even toggling camera button

asyfsaeed:
seems like trackUnpublished causing issue when pressed frequently. any help

timmydoza:
Thanks for the additional information @asyfsaeed!

Unfortunately, I'm still not able to reproduce this issue. However, I will soon have a fix available for a related issue that arises when the 'Mute Video' button is pressed frequrently (see https://github.com/twilio/twilio-video-app-react/issues/218). I'll keep you posted. 

Thanks!

timmydoza:
@asyfsaeed I've merged a fix that may help your issue. Please try it out and let me know how it works for you. Thanks!

PR #240, opened 2020-06-12
Fix video publishing

### LINKED ISSUE #218: Unhandled Rejection (Error): The [LocalVideoTrack #2: b4ea0559-c915-4cd4-b278-cc046adc0c2a] was unpublished
**Describe the bug**
During the video meeting switching on/off video sharing button quickly causes the following error.
`Unhandled Rejection (Error): The [LocalVideoTrack #2: b4ea0559-c915-4cd4-b278-cc046adc0c2a] was unpublished`

Here is the screenshot
Screenshot 2020-05-27 at 6 21 41 PM

P.S. I'm running my code locally.

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-683](https://issues.corp.twilio.com/browse/AHOYAPPS-683)

### Description

This PR changes the logic of the publishing/unpublishing of the local video track in the following ways:

- The Mute Video button is now throttled so that it can only be clicked once every 200ms. This prevents issues caused by clicking the button excessively fast.
- The useLocalVideoToggle hook only publishes/unpublishes the local video track when it is not already doing so.
- If there is a publishing error, an error modal will be displayed instead of the app crashing. 

Resolves #218 and #233 

This PR can be manually tested by entering a room, and then clicking the Mute Video button as quickly as possible. There will be no errors and the button will work as expected. 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/Controls/ToggleVideoButton/ToggleVideoButton.tsx
+++ b/src/components/Controls/ToggleVideoButton/ToggleVideoButton.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useCallback, useRef } from 'react';
 import { createStyles, makeStyles, Theme } from '@material-ui/core/styles';
 
 import Fab from '@material-ui/core/Fab';
@@ -19,6 +19,14 @@ const useStyles = makeStyles((theme: Theme) =>
 export default function ToggleVideoButton(props: { disabled?: boolean }) {
   const classes = useStyles();
   const [isVideoEnabled, toggleVideoEnabled] = useLocalVideoToggle();
+  const lastClickTimeRef = useRef(0);
+
+  const toggleVideo = useCallback(() => {
+    if (Date.now() - lastClickTimeRef.current > 200) {
+      lastClickTimeRef.current = Date.now();
+      toggleVideoEnabled();
+    }
+  }, [toggleVideoEnabled]);
 
   return (
     <Tooltip
@@ -26,7 +34,7 @@ export default function ToggleVideoButton(props: { disabled?: boolean }) {
       placement="top"
       PopperProps={{ disablePortal: true }}
     >
-      <Fab className={classes.fab} onClick={toggleVideoEnabled} disabled={props.disabled}>
+      <Fab className={classes.fab} onClick={toggleVideo} disabled={props.disabled}>
         {isVideoEnabled ? <Videocam /> : <VideocamOff />}
       </Fab>
     </Tooltip>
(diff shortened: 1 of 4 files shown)


src/components/Controls/ToggleVideoButton/ToggleVideoButton.tsx
```diff
@@ -19,14 +19,22 @@ const useStyles = makeStyles((theme: Theme) =>
 export default function ToggleVideoButton(props: { disabled?: boolean }) {
   const classes = useStyles();
   const [isVideoEnabled, toggleVideoEnabled] = useLocalVideoToggle();
+  const lastClickTimeRef = useRef(0);
+
+  const toggleVideo = useCallback(() => {
+    if (Date.now() - lastClickTimeRef.current > 200) {
```
charliesantos, line 25: I think we should use `lodash.throttle`. Same way how you did it here https://github.com/twilio/twilio-video-app-react/blob/master/src/components/Controls/useIsUserActive/useIsUserActive.ts

timmydoza: It isn't really possible to use `throttle` in this instance. For example, this doesn't work:

```
export default function ToggleVideoButton(props: { disabled?: boolean }) {
  const classes = useStyles();
  const [isVideoEnabled, toggleVideoEnabled] = useLocalVideoToggle();

  const toggleVideo = useCallback(throttle(toggleVideoEnabled, 200), [toggleVideoEnabled]);

  return (
    <Tooltip
      title={isVideoEnabled ? 'Mute Video' : 'Unmute Video'}
      placement="top"
      PopperProps={{ disablePortal: true }}
    >
      <Fab className={classes.fab} onClick={toggleVideo} disabled={props.disabled}>
        {isVideoEnabled ? <Videocam /> : <VideocamOff />}
      </Fab>
    </Tooltip>
  );
}
```

That's because calling `toggleVideoEnabled` will cause `useCallback` to produce a new function (because the `useLocalVideoToggle` creates new `isVideoEnabled` and `toggleVideoEnabled` variables whenever the local video is toggled). So `throttle` ends up being called everytime the button is clicked, which means that the throttle effect is lost because `onClick` calls a new function each time instead of the same function as before.

The example you provided works fine because `useEffect` runs only once (note the empty dependency array). There are some `useThrottle` packages that can be installed from NPM, but I found it easy enough just to do it myself. If we need to add this functionality to other components, then I'll can add my own `useThrottle` hook to the codebase.

charliesantos: Ok 😞

PR #241, opened 2020-06-12
Ahoyapps 515 unsupported browsers

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-515](https://issues.corp.twilio.com/browse/AHOYAPPS-515)

### Description

This PR adds a useful error message if the `isSupported` boolean is `true`. 

`isSupported` is always false when `window.isSecureContext` is false. In such cases we display the following text:

image

In cases where the browser is not supported, but the context is secure, the following error is displayed:

image

The link takes the user to our documentation: https://www.twilio.com/docs/video/javascript#supported-browsers

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/UnsupportedBrowserWarning/UnsupportedBrowserWarning.tsx
+++ b/src/components/UnsupportedBrowserWarning/UnsupportedBrowserWarning.tsx
@@ -1,0 +1,65 @@
+import React from 'react';
+import Video from 'twilio-video';
+import { Container, Link, Typography, Paper, Grid } from '@material-ui/core';
+import { makeStyles } from '@material-ui/core/styles';
+
+const useStyles = makeStyles({
+  container: {
+    marginTop: '2.5em',
+  },
+  paper: {
+    padding: '1em',
+  },
+  heading: {
+    marginBottom: '0.4em',
+  },
+});
+
+export default function({ children }: { children: React.ReactElement }) {
+  const classes = useStyles();
+
+  if (!Video.isSupported) {
+    return (
+      <Container>
+        <Grid container justify="center" className={classes.container}>
+          <Grid item xs={12} sm={6}>
+            <Paper className={classes.paper}>
+              <Typography variant="h4" className={classes.heading}>
+                Browser not supported
+              </Typography>
+              <Typography>
+                This is not a supported browser. Please refer to our list of{' '}
+                <Link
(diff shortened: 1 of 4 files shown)


src/components/UnsupportedBrowserWarning/UnsupportedBrowserWarning.tsx
```diff
@@ -0,0 +1,65 @@
...
+                <Link
+                  href="https://www.twilio.com/docs/video/javascript#supported-browsers"
+                  target="_blank"
+                  rel="noopener"
+                >
+                  supported browsers
+                </Link>
+                .
+                <br />
+                {!window.isSecureContext && (
```
charliesantos, line 41: @timmydoza @manjeshbhargav should this be detected in the SDK, and not in the consuming app?

timmydoza: It is detected by the SDK but in an indirect way. `navigator.mediaDevices` isn't present in insecure contexts, and `isSupported` is false when `navigator.mediaDevices` isn't present. 

I just wanted to leave a message for the user in insecure contexts because it seems to be a relatively common problem (users try to access the react dev server over a local network).

charliesantos: I understand the use case. But I was hoping the SDK should be returning this error so that every WebApp that consumes the SDK doesn't have to implement this type of check. Can we confirm with the SDK team?

For background, we also had this issue in VoiceJS so Aymen asked us to put this check in the Voice SDK. It's possible that the Video SDK added this already, but I'm not 100% sure.

timmydoza: I see. I think the Video SDK only returns the `isSupported` boolean. It doesn't throw any errors if the browser is not supported. It seems that the Video SDK leaves it up to the app developer to not call `Video.createLocalTracks()` when `navigator.mediaDevices` is not present. @manjeshbhargav what do you think?

timmydoza:
> @timmydoza I think it is sufficient to just mention both supported browsers and secure contexts when `isSupported` is false.

Thanks @manjeshbhargav! I like the idea - it's simple. I've applied your suggestion:

image

PR #268, opened 2020-07-09
Ahoyapps 734 add track restart feature

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [AHOYAPPS-734](https://issues.corp.twilio.com/browse/AHOYAPPS-734)

### Description

This PR upgrades twilio-video to 2.7.0 and implements the new `track.restart()` feature.  The `track.restart()` feature allows users to change the constraints of their local audio and video tracks without the hassle of unpublishing and republishing the tracks to/from the room.

The `track.restart()` feature replaces publish/unpublish logic in the `AudioInputList`, `VideoInputList`, and `FlipCameraButton` components. 

Additionally, the `VideoTrack` and `AudioLevelIndicator` components had to be updated because they rely on the the `mediaStreamTrack` object of the local tracks. Because the `mediaStreamTrack` can now change without its parent track changing, components must now use the new `useMediaStreamTrack` hook in order to reliabley use that object. 

I've also tested that the `AudioLevelIndicator` component continues to function properly in the following scenarios:
- Switching between tabs in desktop Safari 
- Switching between audio input devices in desktop Firefox
- Multiple instances of the `AudioLevelIndicator` in iOS Safari
- iOS Safari after the microphone has been taken over by another app (Siri)

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18483,9 +18483,8 @@
       }
     },
     "twilio-video": {
-      "version": "2.6.0",
-      "resolved": "https://registry.npmjs.org/twilio-video/-/twilio-video-2.6.0.tgz",
-      "integrity": "sha512-Q/IR7UUc5PTU1uDEzZQuSaeaQH3OOTyw5Xklmq6oqomymVqYDZ6CEuEMHsqxtDajoIOSaflalWYDIQoSE+hS3Q==",
+      "version": "https://github.com/twilio/twilio-video.js/archive/2.7.0-rc2.tar.gz",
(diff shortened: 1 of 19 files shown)


manjeshbhargav:
@timmydoza ,

I think we should have a constant video capture constraints object instead of different constant properties. Also, we need to make sure we capture video from the correct device id when the app is foregrounded.

timmydoza:
> @timmydoza ,
> 
> I think we should have a constant video capture constraints object instead of different constant properties. Also, we need to make sure we capture video from the correct device id when the app is foregrounded.

Thanks! I applied both of these changes.

Issue #262, opened 2020-07-04
Multiple video tracks


bojantv:
**Describe the bug**
Multiple tracks appear for the same video with a different name and sid.

**To Reproduce**
1. Open app
2. Join room
3. Start clicking on the camera button to turn the video on and off repeatedly
4. Observe new video tracks showing up in the publications list and in HTML

**Expected behavior**
No additional video tracks should appear, it should only be a total of one video track.

**Screenshots**
![image](https://user-images.githubusercontent.com/58706872/86521324-215e3900-be47-11ea-9469-5b97f638e725.png)
![image](https://user-images.githubusercontent.com/58706872/86521376-de509580-be47-11ea-98bc-e50da69c39ea.png)

**Environment (please complete the following information):**
 - OS: macOS Catalina
 - Browser: Chrome 83
 - App Version: 0.1.0

**Additional context**

timmydoza:
Thanks for the issue @bojantv!

Unfortunately, I am unable to reproduce the issue that you describe. 
One thing that I noticed, is that the `data-cy-participant` attribute outputs a different value than it should. For me, is just lists the participants identity `data-cy-participant="timmydoza"`. For you it lists a whole object. This difference isn't related to your bug, but it makes me wonder if you are running a different version of the code than me.

Just curious, have you made any modifications to the code base? If you have, is it possible for you to leave a link to your repo? I'd be better able to help you if I could see your modifications.  

Thanks!

bojantv:
Hi, @timmydoza thanks for responding. I manually added that JSON string as identity, different reasons. I was using your app clean, no modifications whatsoever. By just clicking on the video on/video off eventually a double video will appear though they will be layered one on top another so it's very hard to notice it until you check HTML. If you switch from absolute to relative position the other video would appear underneath. Give it another go, please. It takes 10-20 clicks, I'm sure the other video will appear.
I've just reproduced it again clean the latest code, "test" as identity. Just go wild on that camera button (obviously we don't use it like that but in some instances, id does happen with normal usage)

![image](https://user-images.githubusercontent.com/58706872/87034705-3557c100-c1e0-11ea-9201-ddcc90647e74.png)

timmydoza:
Thanks for the info @bojantv!

Looks like I can reproduce the issue. I have to click that button as fast as I can for a few seconds before I see the issue. I'll take a look when I get a chance, but this issue will get lower priority since it only seems to occur with more extreme usage of the Mute Video button. 

[...]

PR #336, opened 2020-10-21
Increase throttle interval in ToggleVideoButton component

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- NA

### Description

This PR increases the throttle interval in the ToggleVideoButton component. This should help to solve the issue described in https://github.com/twilio/twilio-video-app-react/issues/262

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/Buttons/ToggleVideoButton/ToggleVideoButton.test.tsx
+++ b/src/components/Buttons/ToggleVideoButton/ToggleVideoButton.test.tsx
@@ -55,9 +55,9 @@ describe('the ToggleVideoButton component', () => {
     const wrapper = shallow(<ToggleVideoButton />);
     Date.now = () => 100000;
     wrapper.simulate('click'); // Should register
-    Date.now = () => 100100;
+    Date.now = () => 100500;
     wrapper.simulate('click'); // Should be ignored
-    Date.now = () => 100300;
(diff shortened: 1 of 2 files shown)


charliesantos (changes requested):
Maybe there's a cleaner way to solve this. Is there a way to check if the video has been turned on/off successfully then update the state? That way, we don't have to rely on throttling button clicks.

timmydoza:
> Maybe there's a cleaner way to solve this. Is there a way to check if the video has been turned on/off successfully then update the state? That way, we don't have to rely on throttling button clicks.

There is a check in `useLocalVideoToggle` that prevents anything from happening while a camera track is being acquired and published. So nothing will break even if track acquisition & publishing takes longer than the 500ms throttle interval in the button. 

https://github.com/twilio/twilio-video-app-react/blob/master/src/hooks/useLocalVideoToggle/useLocalVideoToggle.tsx#L18

Still, for some reason, issues can arise when you repeatedly hit the Mute Video button quickly for a few minutes. Increasing the throttle interval is only meant to prevent this rare edge case.

charliesantos (approved):

PR #376, opened 2020-12-02
Ensure dimension changes invalidate view

tomhicks:
The track.dimensions object remains the same throughout the lifespan of the track, but its values are updated. This means that the setter always receives the same object and therefore doesn't result in the state actually changing.

By copying the dimension values into a new object, we ensure that consumers of this hook are always kept up to date with the new dimensions.

**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [AHOYAPPS-0000](https://issues.corp.twilio.com/browse/AHOYAPPS-0000)

### Description

A description of what this PR does.

### Before review
* [ ] Updated CHANGELOG.md if necessary
* [ ] Added unit tests if necessary
* [ ] Updated affected documentation
* [ ] Verified locally with `npm test`
* [ ] Manually sanity tested running locally
* [ ] Included screenshot as PR comment (if needed)
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

PR #387, opened 2020-12-16
Update useDimensionsChanged hook so that it causes re-render

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This pull request contains a bugfix submitted by Github user @tomhicks. I've pulled his changes into a new PR so that the Circle CI tests can run on it. Here is the original PR: https://github.com/twilio/twilio-video-app-react/pull/376

This PR fixes an issue found in the `useDimensionsChanged` hook, where the `handleDimensionsChanged` function would set the same object reference as the one already stored in the hook's state. This means that component that use this hook would not re-render when a track's dimensions have changed. 

`handleDimensionsChanged` now returns a new object reference when a track's dimensions have changed, which will correctly cause components to re-render.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/hooks/useVideoTrackDimensions/useVideoTrackDimensions.tsx
+++ b/src/hooks/useVideoTrackDimensions/useVideoTrackDimensions.tsx
@@ -10,7 +10,10 @@ export default function useVideoTrackDimensions(track?: TrackType) {
     setDimensions(track?.dimensions);
 
     if (track) {
-      const handleDimensionsChanged = (track: TrackType) => setDimensions(track?.dimensions);
+      const handleDimensionsChanged = (track: TrackType) => setDimensions({
+        width: track.dimensions.width,
+        height: track.dimensions.height
+      });
       track.on('dimensionsChanged', handleDimensionsChanged);
       return () => {
         track.off('dimensionsChanged', handleDimensionsChanged);
--- a/src/hooks/useVideoTrackDimensions/useVideoTrackDimensions.test.tsx
+++ b/src/hooks/useVideoTrackDimensions/useVideoTrackDimensions.test.tsx
@@ -30,6 +30,20 @@ describe('the useVideoTrackDimensions hook', () => {
     expect(result.current).toEqual({ height: 300, width: 600 });
   });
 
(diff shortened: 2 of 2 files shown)


src/hooks/useVideoTrackDimensions/useVideoTrackDimensions.tsx
```diff
@@ -10,7 +10,10 @@ export default function useVideoTrackDimensions(track?: TrackType) {
     setDimensions(track?.dimensions);
 
     if (track) {
-      const handleDimensionsChanged = (track: TrackType) => setDimensions(track?.dimensions);
+      const handleDimensionsChanged = (track: TrackType) => setDimensions({
```
mhuynh5757, line 13: Do we still need the `?` postfix operator on `track` here? We could also do something like `setDimensions({ ...track?.dimensions })` so we get all of the properties like we used to, just in case something relied on any other `track` properties.

tomhicks: If you need the null coaslescing, the types are wrong, so change those rather than reinstating what appears to be a redundant operator.

You're probably right about cloning the track dimensions to keep other properties, although you might have other deeply-nested objects that will suffer from the 'same reference' problem.

It might be better to specify exactly the return shape of this hook, rather than have it infer it from elsewhere.

timmydoza: > Do we still need the ? postfix operator on track here?

No - it isn't needed inside the `if (track) {` block.

```
if (track) {
  // track will never be undefined in here
}
```

> We could also do something like setDimensions({ ...track?.dimensions }) so we get all of the properties like we used to, just in case something relied on any other track properties.

We could, but `width` and `height` are the only properties on this object, so I think the code is fine as is. https://media.twiliocdn.com/sdk/js/video/releases/2.10.0/docs/VideoTrack.html#.Dimensions

mhuynh5757 (approved):

PR #402, opened 2021-01-20
Fix sidebar flicker

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This PR fixes the flicker that is sometimes seen in the participant sidebar when the browser window just short enough to make the scrollbar appear. 

Before:
Flicker

After:
noflicker

The flicker is caused by a CSS infinite loop. When the scrollbar appears in the sidebar, it reduces the width of the participant thumbnails, which reduces their height (to preserve aspect ratio), which reduces the length of the list, which removes the scroll bar, which increases the width of the participant thumbnails, which increases their length, and so on.

This fix adds an additional div in the sidebar (the 'innerScrollContainer'). Now, when the scrollbar appears, it reduces the size of the innerScrollContainer and not the participant thumbnails. 

Note: To reproduce this issue on a Mac laptop, you must have an external mouse connected. Having an external mouse connected can change the styling of the scrollbars that appear. 

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/ParticipantList/ParticipantList.tsx
+++ b/src/components/ParticipantList/ParticipantList.tsx
@@ -11,7 +11,6 @@ import useScreenShareParticipant from '../../hooks/useScreenShareParticipant/use
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     container: {
-      padding: '2em',
       overflowY: 'auto',
       background: 'rgb(79, 83, 85)',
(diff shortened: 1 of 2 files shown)


charliesantos (approved):
Nice screenshots, explaining the issue clearly 👍

PR #409, opened 2021-01-29
Upgrade twilio-video to 2.11.0

timmydoza:
Remove @types/twilio-video
Update types.ts file

**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This PR upgrades twilio-video.js to version 2.11.0. It also removes the @types/twilio-video type definitions as type definitions for twilio-video are now included starting in version 2.11.0.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/types.ts
```diff
@@ -2,33 +2,12 @@ import { LocalVideoTrack, RemoteVideoTrack, TwilioError } from 'twilio-video';
...
-
-  interface VideoCodecSettings {
-    simulcast?: boolean;
-  }
-
-  interface LocalTrackPublication {
-    setPriority: (priority: Track.Priority) => void;
-  }
-
+  // These help to create union types between Local and Remote VideoTracks
```
charliesantos, line 5: Should we add this in the sdk that way we can get rid of this declaration?

timmydoza: I'm not sure. I'm leaning towards no. This declaration is added because we are creating a union type with LocalVideoTracks and RemoteVideoTracks. Not every application developer may want to do this, so I don't think it needs to be added to the SDK. We aren't adding it here because of a flaw in the SDK types. 

Plus, it isn't taking much space here in `types.js`. I'm not in a hurry to get rid of it 🙂

charliesantos (approved):

PR #433, opened 2021-02-24
Video 4078 add flip camera button to mobile menu

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-4078](https://issues.corp.twilio.com/browse/VIDEO-4078)

### Description

We removed the `FlipCameraButton` from the `MenuBar` to make room for the new chat button. We decided to change this button into a hook - `useFlipCameraToggle` - so that its functionality could be used in the list of MenuItems in `Menu.tsx`. Note that the new Flip Camera MenuItem only appears on mobile when the feature is supported, and is disabled when camera is turned off. We also included the icons from the designs for all menu items.

**Desktop:**

**Mobile:**

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/hooks/useFlipCameraToggle/useFlipCameraToggle.tsx
```diff
@@ -1,16 +1,14 @@
...
-import useMediaStreamTrack from '../../../hooks/useMediaStreamTrack/useMediaStreamTrack';
-import useVideoContext from '../../../hooks/useVideoContext/useVideoContext';
+import useDevices from '../useDevices/useDevices';
+import useMediaStreamTrack from '../useMediaStreamTrack/useMediaStreamTrack';
+import useVideoContext from '../useVideoContext/useVideoContext';
 
-export default function FlipCameraButton() {
+export default function useFlipCameraToggle() {
   const { localTracks } = useVideoContext();
   const [supportsFacingMode, setSupportsFacingMode] = useState<Boolean | null>(null);
```
timmydoza, line 10: Looks like I wrote this line, but looking at it again I'm not sure why `supportsFacingMode` is of type `Boolean | null`. I think it would make more sense if it was just `Boolean`. 

What do you think about this change? It doesn't change the function of this hook, does it?

```suggestion
  const [supportsFacingMode, setSupportsFacingMode] = useState(false);
```

Would also have to change line 22 to: `    if (currentFacingMode && supportsFacingMode === false) {`

This change results in `flipCameraSupported` being of type `Boolean` instead of `Boolean | null`, which makes more sense to me.

src/components/MenuBar/Menu/Menu.test.tsx
```diff
@@ -50,17 +58,60 @@ describe('the Menu component', () => {
...
+      expect(wrapper.find(FlipCameraIcon).exists()).toBe(true);
+      expect(
+        wrapper
+          .find(MenuItem)
+          .at(0)
+          .prop('disabled')
+      ).toBe(false);
+    });
+
+    it('should render a disabled Flip Camera button when facingMode is supported, but videoTrack is undefined', () => {
```
timmydoza, line 94: Nit: I think the text in the `it(...)` functions can be updated here. The Menu component isn't aware of what `facingMode` or `videoTrack` is. The Menu component only sees `flipCameraDisabled` and `flipCameraSupported`, so I think those are the variables that should be mentioned in the `it(...)` functions (on lines 83 and 109 too).

timmydoza (changes requested):
Looks good!!

I'm noticing that the Figma designs for the menu look different than the Abstract designs. Let's quickly sync up with the designers to figure out which look we should go with.

Abstract:
image

Figma: 
image

src/hooks/useFlipCameraToggle/useFlipCameraToggle.tsx
olipyskoty: that makes sense to me too. i googled around a little bit about `MediaStreamTrack` since it was new to me and was able to access this property locally by exploring `twilioRoom.localParticipant.tracks`. I confirmed that `facingMode` is not included in the object returned by `mediaStreamTrack.getSettings()` when on desktop, which means it's not supported so I agree that we should use `false.` TLDR: I agree and made the update 🙂

timmydoza (approved):
Nice work!

PR #451, opened 2021-03-09
Video 3728 add input field and send button

olipyskoty:
**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [VIDEO-3728](https://issues.corp.twilio.com/browse/VIDEO-3728)

### Description

This PR adds an input field and send button to the chat window so the user can type and send messages to other participants. The send button will be highlighted in blue when anything other than whitespace is typed. Users can press `enter` or click the button to send a message, while `enter + shift` creates a new line. Please note that actually sending a message in this PR will throw an error until the new server is complete and merged. We also decided to keep the blue outline around the `<TextareaAutosize />` when it is focused for accessibility reasons. 

**web:**

**mobile:**

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/ChatWindow/ChatInput/ChatInput.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  it('should call the correct function when send message button is clicked', () => {
+    const wrapper = shallow(<ChatInput conversation={{ sendMessage: mockHandleSendMessage } as any} />);
+    wrapper.find(TextareaAutosize).simulate('change', { target: { value: 'I am a message!!!' } });
+    wrapper
+      .find(SendMessageIcon)
+      .parent()
+      .simulate('click');
+    expect(mockHandleSendMessage).toHaveBeenCalled();
+  });
```
timmydoza, line 55: All these tests look good! But I think we could use some additional tests for the `handlereturnKeyPress` function. We'll need to test that it works correctly for the different values of `isMobile`, `event.key`, and `event.shiftKey`.

CircleCI provides a test coverage report for each PR (which can also be generated locally with `npm run test:ci`), which can be used to see if there are any tests that are missing: https://1225-216070925-gh.circle-artifacts.com/0/coverage/lcov-report/src/components/ChatWindow/ChatInput/ChatInput.tsx.html

src/components/ChatWindow/ChatInput/ChatInput.test.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+  it('should call the correct function when send message button is clicked', () => {
+    const wrapper = shallow(<ChatInput conversation={{ sendMessage: mockHandleSendMessage } as any} />);
+    wrapper.find(TextareaAutosize).simulate('change', { target: { value: 'I am a message!!!' } });
+    wrapper
+      .find(SendMessageIcon)
+      .parent()
+      .simulate('click');
+    expect(mockHandleSendMessage).toHaveBeenCalled();
```
timmydoza, line 54: Even better is:

```suggestion
    expect(mockHandleSendMessage).toHaveBeenCalledWith('I am a message!!!');
```

Also, we could have a test to make sure that the whitespace is being trimmed. We could probably just use this same test for that:

`wrapper.find(TextareaAutosize).simulate('change', { target: { value: '  I am a message!!!  \n   ' } });`

src/components/ChatWindow/ChatInput/ChatInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+import { Conversation } from '@twilio/conversations/lib/conversation';
+
+const useStyles = makeStyles({
+  chatInputContainer: {
+    borderTop: '1px solid #e4e7e9',
+    borderBottom: '1px solid #e4e7e9',
+    padding: '1em 1.2em 1em',
+  },
+  textArea: {
+    paddingTop: '0.2em',
```
timmydoza, line 15: I used this for my screenshots in slack:

```
padding: '0.75em 1em`,
marginTop: `0.4em`
```

But feel free to play with it. `padding: '0.75em 1em` makes the padding identical to the padding on the text inputs you see before you enter a room.

timmydoza (changes requested):

src/components/ChatWindow/ChatInput/ChatInput.tsx
```diff
@@ -0,0 +1,89 @@
...
+    background: '#0263E0',
+    cursor: 'pointer',
+  },
+});
+
+interface InputFieldProps {
+  conversation: Conversation;
+}
+
+export default function InputField({ conversation }: InputFieldProps) {
```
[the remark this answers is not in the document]
olipyskoty: ahh good catch - missed this one 👍

src/components/ChatWindow/ChatInput/ChatInput.tsx
olipyskoty: i think this looks good and i like how it stays consistent with the inputs on the "login" page

olipyskoty:
@timmydoza updating the `<div>` to `<button>` tag in the `<ChatWindowHeader>` now includes the blue outline when tabbing to the `<button>`, or when clicking on it. just fyi - but good for accessibility!

src/components/ChatWindow/ChatWindowHeader/ChatWindowHeader.tsx
```diff
@@ -21,6 +21,8 @@ const useStyles = makeStyles(() =>
     closeChatWindow: {
       cursor: 'pointer',
       display: 'flex',
+      background: 'transparent',
+      border: '0',
```
timmydoza, line 25: Nit: I think adding `padding: '0.4em'` helps a little. It makes the blue border square when it is focused. Previously it was a little rectangular.

image

src/components/ChatWindow/ChatInput/ChatInput.tsx
```diff
@@ -70,20 +73,20 @@ export default function InputField({ conversation }: InputFieldProps) {
...
       />
 
-      <div
+      <button
         className={clsx(classes.sendButton, { [classes.activeSendButton]: isValidMessage })}
         onClick={() => handleSendMessage(messageBody)}
       >
         <SendMessageIcon />
-      </div>
+      </button>
```
timmydoza: Looks good! But now that this is a button, I think we should use the `disabled` prop when the button isn't supposed to be clickable:

```
      <button className={classes.sendButton} onClick={() => handleSendMessage(messageBody)} disabled={!isValidMessage}>
        <SendMessageIcon />
      </button>
```
When `disabled` is true, it tells the browser that the button isn't supposed to be clickable. As a result, the button won't be tabbable when it's disabled. So I think it goes a little further than just changing the styles so that it doesn't look clickable. 

Also, with this change, we can take advantage of the `:disabled` [pseudo-class](https://developer.mozilla.org/en-US/docs/Web/CSS/:disabled) to apply the right styles when it's disabled. This way we don't have to worry about conditionally applying the class ourselves. So the `sendButton` class can look like:

```
  sendButton: {
    height: '40px',
    width: '40px',
    border: '0',
    borderRadius: '4px',
    float: 'right',
    display: 'flex',
    alignItems: 'center',
    justifyContent: 'center',
    marginTop: '1em',
    background: '#0263E0',
    cursor: 'pointer',
    '&:disabled': {
      background: 'none',
      cursor: 'default',
    },
  },
  ```

The `&` is just a substitute for the classname of the parent object. So `'&:disabled'` gets turned into `.sendButton:disabled`.

Then, the test will have to be updated to look for the `disabled` prop instead of the `activeSendButton` class (which we can get rid of). For example:

```
    expect(
      wrapper
        .find(SendMessageIcon)
        .parent()
        .prop('disabled')
    ).toBe(true);
```

timmydoza (changes requested):

src/components/ChatWindow/ChatInput/ChatInput.tsx
olipyskoty: good point! this looks nicer too

timmydoza (approved):

PR #458, opened 2021-03-17
Video 3888 file input for conversations

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-3888](https://issues.corp.twilio.com/browse/VIDEO-3888)
- [VIDEO-3889](https://issues.corp.twilio.com/browse/VIDEO-3889)

### Description

This PR adds support for sending and receiving files while in a room. It uses the Twilio Conversations SDK to send the files in the chat. I left a link to a quick-deploy version of this PR in our slack channel if you want to try it out.  

On Monday when I get back I'll take a look at the failing test. Otherwise I think the code is ready for review.

files

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/ChatWindow/ChatInput/ChatInput.tsx
```diff
@@ -81,9 +126,28 @@ export default function ChatInput({ conversation }: ChatInputProps) {
         value={messageBody}
       />
 
-      <button className={classes.sendButton} onClick={() => handleSendMessage(messageBody)} disabled={!isValidMessage}>
-        <SendMessageIcon />
-      </button>
+      <Grid container alignItems="flex-end" justify="flex-end" wrap="nowrap">
+        <input ref={fileInputRef} type="file" style={{ display: 'none' }} onChange={handleSendFile} />
```
charliesantos, line 130: Can we filter the files here to only allow what Twilio Conversations support? https://www.twilio.com/docs/conversations/conversations-limits#media-limits
I believe you can use the `accept` attribute. Also, add check for max file size. If it's more than what Twilio Conversations support, don't even start uploading.

charliesantos (changes requested):

src/components/ChatWindow/ChatInput/ChatInput.tsx
timmydoza: I think those limits are only relevant to Conversations that send messages to an SMS channel. In Conversations that only have chat participants, any file type can be sent. All file types that I have uploaded in chat have succeeded.

Also, the Conversations SDK already checks the file size and lets us know when it is too large, in which case we display a useful error to the user:

image

Is there a benefit to checking the file size ourselves instead of letting the SDK do it?

charliesantos: If the SDK is already checking it, then I think we're good. How about the file types? Is there a doc somewhere that mentions unsupported/supported files? For example, it's generally not advisable to upload executable files. I'm curious how the SDK or server handles this.

timmydoza: It looks like there are no limits with respect to the file types that can be sent between two chat participants. I've even been able to send executable files in my tests.

So any file type limitations we add would be specific to our app only, and not specific to the Conversations platform. I'm happy to add some (it's technically easy), but I think this is a decision that Sean or Donal should weigh in on. I'll ask around.

timmydoza: @charliesantos I added the `accept` attribute to the file input. I used some MIME types instead of file extensions just to allow for some greater flexibility. Also, using `image/*` allows mobile users to take a picture to upload to the chat. What do you think?

charliesantos: It looks great! Thanks for adding this. 🚢  🚀

charliesantos (approved):

PR #469, opened 2021-03-25
Update MessageList component to change how messages are grouped

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This PR changes how `MessageInfo` components are displayed in the `MessageList` component. Instead of displaying a `MessageInfo` component whenever a message has a new author, `MessageInfo` components are now displayed when a message has a new author or formatted timestamp. 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/ChatWindow/MessageList/MessageList.tsx
+++ b/src/components/ChatWindow/MessageList/MessageList.tsx
@@ -28,11 +28,19 @@ export default function MessageList({ messages }: MessageListProps) {
         const time = message.dateCreated
           .toLocaleTimeString('en-us', { hour: 'numeric', minute: 'numeric' })
           .toLowerCase();
+
+        const previousTime = messages[idx - 1]?.dateCreated
+          .toLocaleTimeString('en-us', { hour: 'numeric', minute: 'numeric' })
+          .toLowerCase();
+
+        // Display the MessageInfo component when the author or formatted timestamp differs from the previous message
+        const shouldDisplayMessageInfo = time !== previousTime || message.author !== messages[idx - 1]?.author;
+
         const isLocalParticipant = localParticipant.identity === message.author;
 
         return (
           <React.Fragment key={message.sid}>
-            {messages[idx - 1]?.author !== message.author && (
+            {shouldDisplayMessageInfo && (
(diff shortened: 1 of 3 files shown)


src/components/ChatWindow/MessageList/MessageList.tsx
```diff
@@ -28,11 +28,19 @@ export default function MessageList({ messages }: MessageListProps) {
         const time = message.dateCreated
           .toLocaleTimeString('en-us', { hour: 'numeric', minute: 'numeric' })
           .toLowerCase();
+
+        const previousTime = messages[idx - 1]?.dateCreated
+          .toLocaleTimeString('en-us', { hour: 'numeric', minute: 'numeric' })
+          .toLowerCase();
```
olipyskoty, line 34: quick question, since `time` and `previousTime` are almost identical, would this be better as a function that accepts a time and we could pass in `message.dateCreated` or `messages[idx-1]?.dateCreated`? or maybe that would just make it more confusing 🤔

src/components/ChatWindow/MessageList/MessageList.tsx
```diff
@@ -28,11 +28,19 @@ export default function MessageList({ messages }: MessageListProps) {
...
+
+        // Display the MessageInfo component when the author or formatted timestamp differs from the previous message
+        const shouldDisplayMessageInfo = time !== previousTime || message.author !== messages[idx - 1]?.author;
```
olipyskoty, line 37: good variable name! 👍

olipyskoty (approved):
looks good to me, just left a quick question

src/components/ChatWindow/MessageList/MessageList.tsx
timmydoza: Good question!

There is some repeated code here. In my opinion, I think it can be fine to have some duplicate code. When code is repeated just once, I sometimes feel that it isn't worth it to create an abstraction in order to get rid of the duplicated code. Sometimes you end up with more code than you started with. 

However, I tried it out in this instance, and with your idea, I found that I was able to both reduce the amount of code in this file and make it more readable at the same time. So here it definitely helps! I'll upload my change.

PR #492, opened 2021-04-14
Add ability to change twilio environments

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

### Description

A description of what this PR does.

### Before review
* [ ] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/server/createExpressHandler.ts
+++ b/server/createExpressHandler.ts
@@ -7,6 +7,7 @@ const { TWILIO_ACCOUNT_SID, TWILIO_API_KEY_SID, TWILIO_API_KEY_SECRET, TWILIO_CO
 
 const twilioClient = Twilio(TWILIO_API_KEY_SID, TWILIO_API_KEY_SECRET, {
   accountSid: TWILIO_ACCOUNT_SID,
+  region: process.env.REACT_APP_TWILIO_ENVIRONMENT,
 });
 
 const context: ServerlessContext = {
--- a/server/createExpressHandler.ts
+++ b/server/createExpressHandler.ts
@@ -3,11 +3,17 @@ import { Request, Response } from 'express';
 import { ServerlessContext, ServerlessFunction } from './types';
 import Twilio from 'twilio';
 
-const { TWILIO_ACCOUNT_SID, TWILIO_API_KEY_SID, TWILIO_API_KEY_SECRET, TWILIO_CONVERSATIONS_SERVICE_SID } = process.env;
+const {
+  TWILIO_ACCOUNT_SID,
+  TWILIO_API_KEY_SID,
+  TWILIO_API_KEY_SECRET,
+  TWILIO_CONVERSATIONS_SERVICE_SID,
+  REACT_APP_TWILIO_ENVIRONMENT,
+} = process.env;
 
 const twilioClient = Twilio(TWILIO_API_KEY_SID, TWILIO_API_KEY_SECRET, {
   accountSid: TWILIO_ACCOUNT_SID,
-  region: process.env.REACT_APP_TWILIO_ENVIRONMENT,
+  region: REACT_APP_TWILIO_ENVIRONMENT === 'prod' ? undefined : REACT_APP_TWILIO_ENVIRONMENT,
 });
 
 const context: ServerlessContext = {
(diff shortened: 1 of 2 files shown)


src/utils/useConnectionOptions/useConnectionOptions.ts
```diff
@@ -37,13 +37,21 @@ export default function useConnectionOptions() {
     // their individual bandwidth constraints. Simulcast should be disabled if
     // you are using Peer-to-Peer or 'Go' Rooms.
     preferredVideoCodecs: [{ codec: 'VP8', simulcast: roomType !== 'peer-to-peer' && roomType !== 'go' }],
+
+    //@ts-ignore - Internal use only. This property is not exposed in type definitions.
+    environment: process.env.REACT_APP_TWILIO_ENVIRONMENT,
```
makarandp0, line 42: we should include this only when `process.env.REACT_APP_TWILIO_ENVIRONMENT !== 'prod` ` so that apps that base off react app wont pass this parameter.

makarandp0 (dismissed):
One change request. LGTM otherwise

src/utils/useConnectionOptions/useConnectionOptions.ts
timmydoza: It seems to work fine when `REACT_APP_TWILIO_ENVIRONMENT=prod`. I'm not sure that this is needed.

Setting `environment: 'prod'` here is the same thing the SDK does [here](https://github.com/twilio/twilio-video.js/blob/master/lib/connect.js#L212). So I don't think it causes any issues.

makarandp0: sure, it does not cause any issues, but its not one of the documented parameter. I would not recommend that our customers pass it.  I would do  something like:

```
  // Internal use only - remove this code block
  if (environment !== 'prod') {
    //@ts-ignore - Internal use only. This property is not exposed in type definitions.
    connectionOptions!.environment = environment;
    if (environment === 'dev') {
      //@ts-ignore - Internal use only. This property is not exposed in type definitions.
      connectionOptions!.wsServer = 'wss://us2.vss.dev.twilio.com/signaling';
    }
  }
```

timmydoza: > I would not recommend that our customers pass it

I agree. I wouldn't customers using `REACT_APP_TWILIO_ENVIRONMENT` either, which is why I decided to leave it out of the [.env.example](https://github.com/twilio/twilio-video-app-react/blob/master/.env.example) file. I think this is enough to keep customers away from it. It will be an undocumented parameter, just like `connectOptions.environment`.

I don't think we need to add another if statement to prevent `environment: 'prod'` from being set, especially since that what the SDK sets it to by default.

makarandp0: sure.

server/createExpressHandler.ts
```diff
@@ -3,10 +3,17 @@ import { Request, Response } from 'express';
...
+  TWILIO_ACCOUNT_SID,
+  TWILIO_API_KEY_SID,
+  TWILIO_API_KEY_SECRET,
+  TWILIO_CONVERSATIONS_SERVICE_SID,
+  REACT_APP_TWILIO_ENVIRONMENT,
+} = process.env;
 
 const twilioClient = Twilio(TWILIO_API_KEY_SID, TWILIO_API_KEY_SECRET, {
   accountSid: TWILIO_ACCOUNT_SID,
+  region: REACT_APP_TWILIO_ENVIRONMENT === 'prod' ? undefined : REACT_APP_TWILIO_ENVIRONMENT,
```
makarandp0, line 16: is `undefined` one of the allowed values?

makarandp0 (approved):

server/createExpressHandler.ts
timmydoza: Yeah. Here `region: 'prod'` breaks everything, but `undefined` is fine when `REACT_APP_TWILIO_ENVIRONMENT === 'prod'`

PR #495, opened 2021-04-16
Video 4885 recording cypress tests

olipyskoty:
**Contributing to Twilio**

- [X] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [VIDEO-4885](https://issues.corp.twilio.com/browse/VIDEO-4885)

### Description

This PR adds cypress tests for the recording start/stop feature.

### Before review
* [X] Verified locally with `npm test`
* [X] Manually sanity tested running locally
* [X] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/cypress/integration/twilio-video.spec.js
+++ b/cypress/integration/twilio-video.spec.js
@@ -102,7 +102,7 @@ context('A video app user', () => {
       it('should see the recording indicator and notification after clicking "Start Recording"', async () => {
         cy.get('[data-cy-recording-indicator]').should('be.visible');
         cy.contains('Recording has started').should('be.visible');
-        await delay(2000)
+        await delay(2000);
         cy.get('[data-cy-more-button]')
           .last()
           .click();
--- a/cypress/integration/twilio-video.spec.js
+++ b/cypress/integration/twilio-video.spec.js
@@ -87,13 +87,35 @@ context('A video app user', () => {
       cy.get('[data-cy-main-participant]').should('contain', 'testuser');
     });
 
-    describe.only('the recording start/stop feature', () => {
+    describe('the recording start/stop feature', () => {
+      before(() => {
+        cy.get('[data-cy-more-button]')
+          .last()
+          .click();
+        cy.get('[data-cy-recording-button]').click();
+      });
+
       after(async () => {
         await delay(3000);
       });
 
-      it('should see the recording indicator after clicking "Start Recording"', () => {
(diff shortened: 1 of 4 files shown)


cypress/integration/twilio-video.spec.js
```diff
@@ -82,6 +82,38 @@ context('A video app user', () => {
       cy.getParticipant('test1').should('not.exist');
       cy.get('[data-cy-main-participant]').should('contain', 'testuser');
     });
+
+    describe('the recording start/stop feature', () => {
+      before(() => {
+        cy.get('[data-cy-more-button]')
+          .last()
```
timmydoza, line 89: Oh I see. `.last()` is needed because there's a "More" button in the top menu bar (on mobile), and the bottom menu bar (on desktop). 

An alternative way to select the right "More" button is to use the ["descendant combinator"](https://developer.mozilla.org/en-US/docs/Web/CSS/Descendant_combinator) to select the button that exists in the `footer` element:

`cy.get('footer [data-cy-more-button]').click()`.

Not sure which is better here. I think `footer [data-cy-more-button]` is a little more specific, and would be better if there was a chance that the two buttons would change order, but there isn't a chance of that happening in this app. I'll let you decide. 🙂

cypress/integration/twilio-video.spec.js
```diff
@@ -82,6 +82,38 @@ context('A video app user', () => {
...
+      });
+
+      after(() => {
+        cy.wait(3000);
+      });
+
+      it('should see the recording indicator and notification after clicking "Start Recording"', () => {
+        cy.get('[data-cy-recording-indicator]').should('be.visible');
+        cy.contains('Recording has started').should('be.visible');
+        cy.wait(2000);
```
timmydoza, line 101: I think this wait statement should go at the end of the `before()` block. Looks like there isn't a delay between recording start (in the before block) and recording stop (in the "it should see 'Recording Complete' notification" test).

timmydoza (changes requested):
Nicely done! Just one tiny change that is needed.

cypress/integration/twilio-video.spec.js
olipyskoty: cool, thanks! i didn't know that you could do that in cypress. i went with your suggestion because i agree that it is more specific, but it will also help me remember that i could use this syntax in the future 👍

timmydoza (approved):

PR #510, opened 2021-05-07
Conditionally render recording rules button

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- N/A

### Description

This PR adds logic so that the recording start/stop button isn't rendered when the room type is peer-to-peer or go. 

Additionally, some of the token request logic in src/state was modified so that the roomType is saved no matter which authentication scheme is used. 

If the roomType is ever unknown (a developer using their own server that isn't properly set up), then the recording button will be rendered. If the room type doesn't allow recording, the user will see a useful error message:

image

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/MenuBar/Menu/Menu.test.tsx
+++ b/src/components/MenuBar/Menu/Menu.test.tsx
@@ -33,7 +33,11 @@ describe('the Menu component', () => {
 
   beforeAll(() => {
     mockUpdateRecordingRules = jest.fn(() => Promise.resolve());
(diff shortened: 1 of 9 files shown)


olipyskoty (approved):
This looks good to me 👍. Confirmed I could record the room in a `group` room, and did not see the recording option in a `p2p` room.

timmydoza:
> This looks good to me 👍. Confirmed I could record the room in a `group` room, and did not see the recording option in a `p2p` room.

@olipyskoty Thanks for trying it out!

PR #550, opened 2021-07-10
VIDEO-5734/blur background feature

gabriel-jt:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-5734](https://issues.corp.twilio.com/browse/VIDEO-5734)

### Description

Added blur background feature with unit tests

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +67,68 @@ export const backgroundConfig = {
...
+
+  const updateBackgroundSettings = useCallback(
+    async (settings: BackgroundSettings, reset?: boolean, force?: boolean, newRoom?: Room) => {
+      // reset, used for resetting background settings when camera is off/no video tracks
+      let currentRoom = room;
+      if (reset) {
+        updateSettings({ type: 'none' });
+        return;
+      }
+      currentRoom = currentRoom || newRoom;
```
charliesantos, line 98: I wonder if it's possible to get the current room without passing a `room` or `newRoom` param..

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +67,68 @@ export const backgroundConfig = {
...
+      if (!currentRoom) {
+        return;
+      }
+
+      const videoTrackValue = Array.from(currentRoom.localParticipant.videoTracks.values())[0];
```
charliesantos, line 103: Do we have hooks to get the local video track? I only did this for prototype, sorry not meant to be copied.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +67,68 @@ export const backgroundConfig = {
   images,
 };
 
-// TODO : Add video processing logic after backgroundSettings change
-// useEffect hooks, etc ...
+const bgLibSettings = {
+  assetsPath: '/virtualbackground',
```
charliesantos, line 71: Since there's only one property in this object, I think we can just set a constant value for `assetsPath`.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +67,68 @@ export const backgroundConfig = {
...
+        return;
+      }
+
+      const videoTrackValue = Array.from(currentRoom.localParticipant.videoTracks.values())[0];
+      if (!videoTrackValue) {
+        return;
+      }
+      const videoTrack = videoTrackValue.track;
+      const { type } = settings;
+      if (type !== backgroundSettings.type || force) {
```
charliesantos, line 109: Why do we need `force` again?

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +67,68 @@ export const backgroundConfig = {
...
+        if (type === 'none' && videoTrack.processor) {
+          videoTrack.removeProcessor(videoTrack.processor);
+        } else if (type === 'blur') {
+          if (!blurProcessor) {
+            blurProcessor = new GaussianBlurBackgroundProcessor({ ...bgLibSettings });
+            await blurProcessor.loadModel();
+          }
```
charliesantos, line 116: We should probably put this outside of this method.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
charliesantos: There's also a lot of duplicated checks in this if block. I understand you are using my prototype as a reference 😅 . I think this block can be simplified.

Move the initialization outside of updateBackgroundSettings. You might need to wrap it in an useEffect.
```ts
if (!blurProcessor) {
  blurProcessor = new GaussianBlurBackgroundProcessor({ ...bgLibSettings });
  await blurProcessor.loadModel();
}
```

Then you can shrink your if block to something like this
```ts
if (type !== backgroundSettings.type || force) {
  if (videoTrack.processor) {
    videoTrack.removeProcessor(videoTrack.processor);
  }
  if (type === 'blur') {
    videoTrack.addProcessor(blurProcessor);
  }
}
```

charliesantos (changes requested):
Nice one @gabriel-jt ! I did a first pass and left a few comments.

src/hooks/useLocalVideoToggle/useLocalVideoToggle.tsx
```diff
@@ -20,10 +28,22 @@ export default function useLocalVideoToggle() {
         getLocalVideoTrack()
           .then((track: LocalVideoTrack) => localParticipant?.publishTrack(track, { priority: 'low' }))
           .catch(onError)
-          .finally(() => setIspublishing(false));
+          .finally(() => {
+            updateBackgroundSettings(backgroundSettings, false, true);
```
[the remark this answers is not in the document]
timmydoza: I think this could be done in a `useEffect` hook inside of `useBackgroundSettings.tsx`.

timmydoza (changes requested):
Nice work @gabriel-jt! It's fun to try out the blur feature in action. Feel free to ping me if you have any questions about my comments.

package.json
```diff
@@ -6,8 +6,8 @@
   "dependencies": {
     "@material-ui/core": "^4.9.1",
     "@material-ui/icons": "^4.9.1",
-    "@twilio-labs/plugin-rtc": "^0.8.2",
-    "@twilio/conversations": "^1.1.0",
+    "@twilio-labs/plugin-rtc": "0.8.2",
+    "@twilio/conversations": "1.1.0",
     "@twilio/video-processors": "1.0.0",
```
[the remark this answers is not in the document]
charliesantos: @timmydoza , that is correct. The increase in size is due to video processors dependencies. But we will shrink this down significantly in the upcoming release because we will be removing tfjs support. So I think we can leave this as is for now.

package.json
```diff
@@ -6,8 +6,8 @@
   "dependencies": {
     "@material-ui/core": "^4.9.1",
     "@material-ui/icons": "^4.9.1",
-    "@twilio-labs/plugin-rtc": "^0.8.2",
-    "@twilio/conversations": "^1.1.0",
+    "@twilio-labs/plugin-rtc": "0.8.2",
```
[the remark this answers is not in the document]
gabriel-jt: The carats were removed since there were some breaking changes in the newer versions of `@twilio-labs/plugin-rtc` and `@twilio/conversations` that prevented the app from running

src/components/Buttons/EndCallButton/EndCallButton.tsx
```diff
@@ -20,10 +20,15 @@ const useStyles = makeStyles((theme: Theme) =>
 
 export default function EndCallButton(props: { className?: string }) {
   const classes = useStyles();
-  const { room } = useVideoContext();
+  const { room, updateBackgroundSettings } = useVideoContext();
+
+  const handleDisconnect = () => {
+    updateBackgroundSettings({ type: 'none', index: 0 });
```
charliesantos, line 26: I don't think we want to update the settings to 'none' after disconnecting. We want to keep what the user had set right?

src/constants.ts
```diff
@@ -8,3 +8,6 @@ export const DEFAULT_VIDEO_CONSTRAINTS: MediaStreamConstraints['video'] = {
 export const SELECTED_AUDIO_INPUT_KEY = 'TwilioVideoApp-selectedAudioInput';
 export const SELECTED_AUDIO_OUTPUT_KEY = 'TwilioVideoApp-selectedAudioOutput';
 export const SELECTED_VIDEO_INPUT_KEY = 'TwilioVideoApp-selectedVideoInput';
+
+// This is used to store the current background settings in localStorage
+export const BG_SETTINGS_KEY = 'bgsettings5';
```
charliesantos, line 13: Just use `bgsettings` instead of `bgsettings5` 😄 
I put 5 in there in the prototype code because I wanted to bust the cache.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+    setBackgroundSettings(settings);
+    window.localStorage.setItem(BG_SETTINGS_KEY, JSON.stringify(settings));
+  };
+
+  const updateBackgroundSettings = useCallback(
+    async (settings: BackgroundSettings) => {
+      // check if new background settings have been selected
+      if (settings.type !== backgroundSettings.type) {
+        // remove processor
+        if (videoTrack.processor) {
```
charliesantos, line 113: Just tested and this throws an error if I disable my video. Maybe update this to `if (videoTrack && videoTrack.processor)`?

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+  }, []);
+
+  // for reapplying background replacement after camera is toggled on again
+  useEffect(() => {
+    const storedBackgroundSettings = window.localStorage.getItem(BG_SETTINGS_KEY);
+    if (videoTrack && storedBackgroundSettings) {
+      const parsedSettings = JSON.parse(storedBackgroundSettings!);
+      if (parsedSettings.type === 'blur') {
+        videoTrack.addProcessor(blurProcessor);
+      }
```
charliesantos, line 98: Should we just call setBackgroundSettings here? That way, all properties are handled.

```ts
setBackgroundSettings(parsedSettings);
```

charliesantos (changes requested):
Looking really great @gabriel-jt !

package.json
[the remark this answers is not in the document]
timmydoza: Interesting. What errors were you seeing?

The version numbers of `@twilio-labs/plugin-rtc` and `@twilio/conversations` as specified in the package-lock.json file are 0.8.2 and 1.1.0, respectively. package-lock.json dictates the exact versions that should be installed. So I'm wondering if you ran into some other kind of issue, because 
npm shouldn't have installed newer versions. 

I consider it a best practice to leave the carets in (with exception of the react-scripts package). The version numbers listed in package.json describe the range of versions of a package that will be compatible with the app. Note that the package.json file does not dictate which specific version is installed (this is what package-lock.json is for). The caret means that this app can use any package version that has a minor or patch version number (but not major version number) higher than what's specified. The three numbers in a version all have a specific meaning, which is called [semantic versioning](https://docs.npmjs.com/about-semantic-versioning). Generally speaking, any new package version that only increments the minor or patch version numbers should not introduce any breaking changes. 

The `package-lock.json` file serves a different purpose. This file dictates the exact package versions that should be installed. This is to ensure that each time anybody runs `npm install`, the exact same set of packages is installed. This helps reduce weird issues where some users experience bugs that other users don't because they installed a slightly different set of packages. 

So the caret won't introduce any breaking changes, and it also is meant to indicate that newer minor and patch versions can be installed. So let's keep the carets in (and also add it to rimraf, copyfiles, twilio, and @twilio/video-processors). Let me know if there are any issues that you run into. I'd be more than happy to take a look. 🙂 

Also, this [semver calculator](https://semver.npmjs.com/) is a fun way to learn about npm's version syntax.

package.json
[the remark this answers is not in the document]
timmydoza: Cool, sounds good!

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+        if (settings.type === 'blur') {
+          videoTrack.addProcessor(blurProcessor);
+        } else if (settings.type === 'image') {
+          // TODO add image replacement logic
+        }
+        updateSettings(settings);
+      }
+    },
+    [backgroundSettings, videoTrack]
+  );
```
timmydoza, line 125: Really nice refactor here!! 🎉 This function is super easy to make sense of.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+    if (videoTrack && storedBackgroundSettings) {
+      const parsedSettings = JSON.parse(storedBackgroundSettings!);
+      if (parsedSettings.type === 'blur') {
+        videoTrack.addProcessor(blurProcessor);
+      }
+    }
+  }, [videoTrack]);
+
+  // update the backgroundSettings and save into localStorage
+  const updateSettings = (settings: BackgroundSettings): void => {
```
timmydoza, line 103: I think this function would be useful if it were used in many places, but it's only called from `updateBackgroundSettings`. I think we could get rid of it and just move its contents into `updateBackgroundSettings`.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+    return localStorageSettings ? JSON.parse(localStorageSettings) : { type: 'none', index: 0 };
+  });
+
+  // load processors on initialization
+  useEffect(() => {
+    if (!blurProcessor) {
+      blurProcessor = new GaussianBlurBackgroundProcessor({
+        assetsPath: virtualBackgroundAssets,
+      });
+      blurProcessor.loadModel();
```
timmydoza, line 87: I'm not sure that this needs to be in a useEffect hook. Would it work if we moved it outside of `useBackgroundSettings`?

src/constants.ts
timmydoza: What about `export const SELECTED_BACKGROUND_SETTINGS_KEY = 'TwilioVideoApp-selectedBackgroundSettings'`? Just to be more descriptive and to have some consistency with the other keys. With a more descriptive variable name, we won't need a comment to describe the variable.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,61 @@ export const backgroundConfig = {
...
+    }
+  }, []);
+
+  // for reapplying background replacement after camera is toggled on again
+  useEffect(() => {
+    const storedBackgroundSettings = window.localStorage.getItem(BG_SETTINGS_KEY);
+    if (videoTrack && storedBackgroundSettings) {
+      const parsedSettings = JSON.parse(storedBackgroundSettings!);
+      if (parsedSettings.type === 'blur') {
+        videoTrack.addProcessor(blurProcessor);
```
timmydoza, line 97: This seems duplicative of what is already in `updateBackgroundSettings`. Also, I don't think we need to read from local storage every time the user enables their video. We can just rely on the `backgroundSettings` variable, which gets its initial value from local storage. 

After thinking about this hook some, it seems like we just need to run the logic in `updateBackgroundSettings` whenever the settings are updated, or whenever the video track changes. In my mind this calls for a useEffect hook. The only state here is the selected background settings, and the video track. The application of the processor, and saving the settings to local storage are just side-effects of the changes in state. So I think all the code inside of `updateBackgroundSettings` can go in a useEffect hook instead, which will be set up to react to any changes in `backgroundSettings` or `videoTrack`. Instead of passing `updateBackgroundSettings` to the videoProvider, we could just pass `setBackgroundSettings` instead.

What do you think about a hook like this (it incorporates changes from my other comments):

```
let blurProcessor = new GaussianBlurBackgroundProcessor({
  assetsPath: '/virtualbackground',
});
blurProcessor.loadModel();

export default function useBackgroundSettings(videoTrack: LocalVideoTrack) {
  const [backgroundSettings, setBackgroundSettings] = useState<BackgroundSettings>(() => {
    const localStorageSettings = window.localStorage.getItem(BG_SETTINGS_KEY);
    return localStorageSettings ? JSON.parse(localStorageSettings) : { type: 'none', index: 0 };
  });

  useEffect(() => {
    if (videoTrack) {
      if (videoTrack.processor) {
        videoTrack.removeProcessor(videoTrack.processor);
      }
      if (backgroundSettings.type === 'blur') {
        videoTrack.addProcessor(blurProcessor);
      } else if (backgroundSettings.type === 'image') {
        // TODO add image replacement logic
      }
    }
    window.localStorage.setItem(BG_SETTINGS_KEY, JSON.stringify(backgroundSettings));
  }, [videoTrack, backgroundSettings]);

  return [backgroundSettings, setBackgroundSettings] as const;
}
```

src/components/VideoProvider/index.tsx
```diff
@@ -81,7 +81,8 @@ export function VideoProvider({ options, children, onError = () => {} }: VideoPr
   useRestartAudioTrackOnDeviceChange(localTracks);
 
   const [isBackgroundSelectionOpen, setIsBackgroundSelectionOpen] = useState(false);
-  const [backgroundSettings, setBackgroundSettings] = useBackgroundSettings();
+  const videoTracks = localTracks.find(track => track.name.includes('camera')) as LocalVideoTrack;
```
timmydoza, line 84: Nice idea to grab the camera track from here! One small thing: it's possible that `localTracks.find` returns `undefined`, so I think the type of this variable needs to be adjusted. 

`const videoTracks = localTracks.find(track => track.name.includes('camera')) as LocalVideoTrack | undefined;`

timmydoza (changes requested):
Great job refactoring this! I think we can go a little further though. Let me know if you have any questions about anything.

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
gabriel-jt: Hi Tim, after thinking about how to simplify this hook, I managed to simplify its implementation down to something similar to this! After some testing, this also takes care of an important edge case (reapplying the previously selected background settings after a user turns off then turns on their camera)

package.json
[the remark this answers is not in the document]
gabriel-jt: It's been a while since Charlie and I dealt with these issues, so I don't remember them off the top of my head. Thinking back, I think I got those errors after updating all my npm packages, which introduced the breaking changes from the newer versions. I was able to add back the carats with no issues though!

src/components/VideoProvider/index.tsx
```diff
@@ -81,7 +82,8 @@ export function VideoProvider({ options, children, onError = () => {} }: VideoPr
   useRestartAudioTrackOnDeviceChange(localTracks);
 
   const [isBackgroundSelectionOpen, setIsBackgroundSelectionOpen] = useState(false);
-  const [backgroundSettings, setBackgroundSettings] = useBackgroundSettings();
+  const videoTracks = localTracks.find(track => track.name.includes('camera')) as LocalVideoTrack | undefined;
```
charliesantos, line 85:
```suggestion
  const videoTrack = localTracks.find(track => track.name.includes('camera')) as LocalVideoTrack | undefined;
```

src/components/VideoProvider/index.tsx
```diff
@@ -81,7 +82,8 @@ export function VideoProvider({ options, children, onError = () => {} }: VideoPr
...
+  const [backgroundSettings, setBackgroundSettings, removeProcessor] = useBackgroundSettings(videoTracks);
```
charliesantos, line 86:
```suggestion
  const videoTrack = localTracks.find(track => track.name.includes('camera')) as LocalVideoTrack | undefined;
  const [backgroundSettings, setBackgroundSettings, removeProcessor] = useBackgroundSettings(videoTrack);
```

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,40 @@ export const backgroundConfig = {
...
+});
+blurProcessor.loadModel();
+
+export default function useBackgroundSettings(videoTrack: LocalVideoTrack | undefined) {
+  const [backgroundSettings, setBackgroundSettings] = useState<BackgroundSettings>(() => {
+    const localStorageSettings = window.localStorage.getItem(SELECTED_BACKGROUND_SETTINGS_KEY);
+    return localStorageSettings ? JSON.parse(localStorageSettings) : { type: 'none', index: 0 };
+  });
+
+  // Specifically used for handling room disconnection
```
charliesantos, line 85: You can remove this comment. You can reuse this function below.
```suggestion

```

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts
```diff
@@ -65,13 +68,40 @@ export const backgroundConfig = {
...
+    if (videoTrack && videoTrack.processor) {
+      videoTrack.removeProcessor(videoTrack.processor);
+    }
+  };
+
+  useEffect(() => {
+    if (videoTrack) {
+      if (videoTrack.processor) {
+        videoTrack.removeProcessor(videoTrack.processor);
+      }
```
charliesantos, line 96: You can reuse the method you created above
```suggestion
     removeProcessor();
```

gabriel-jt: Hey Charlie, I made it a separate function and didn't use it inside the useEffect hook since I was getting warnings about missing `removeProcessor` as a dependency, but when I include it, I get another warning about how `removeProcessor` is updated on every render. What would be the best step/approach to this?

gabriel-jt: cc @timmydoza

timmydoza: There's a way to make this work, but I wonder if we need to worry about removing the processor when we disconnect from a room?  When we disconnect from a room, track.stop() is automatically called because of this hook [here](https://github.com/twilio/twilio-video-app-react/blob/master/src/components/VideoProvider/useHandleRoomDisconnection/useHandleRoomDisconnection.ts#L22) (this is to turn off the camera light on laptops when the user disconnects from a room).  Then when the user goes back to the Join Room screens, new tracks are acquired.

@charliesantos Since we are stopping the tracks on disconnect, do we need to remove the processor as well?

If we do, I think we can just handle that in the `removeVideoTrack` function in the `useLocalTracks` hook [here](https://github.com/twilio/twilio-video-app-react/blob/master/src/components/VideoProvider/useLocalTracks/useLocalTracks.ts#L45). Something like this maybe (but only if needed):

```
  const removeLocalVideoTrack = useCallback(() => {
    if (videoTrack) {
      if (videoTrack.processor) {
        videoTrack.removeProcessor(videoTrack.processor)
      }
      videoTrack.stop();
      setVideoTrack(undefined);
    }
  }, [videoTrack]);
  ```

Comment (charliesantos) [reply src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.ts:96]: Yes we need to remove the processor that way it won't show in the join screen. On @timmydoza 's suggestion, it doesn't seem to save us anything 😄 . I was hoping we don't have to duplicate the code. I think I'm leaning more towards keeping our current implementation. We will probably remove this later if we decide to launch the virtual background settings UI in the join screen page.

File: src/components/VideoProvider/index.tsx [hunk 22]
```diff
@@ -79,9 +80,12 @@ export function VideoProvider({ options, children, onError = () => {} }: VideoPr
   );
   useHandleTrackPublicationFailed(room, onError);
   useRestartAudioTrackOnDeviceChange(localTracks);
+  console.log('tracks: ');
+  console.log(localTracks);
```
charliesantos, line 84: Please remove

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+const defaultSettings = {
+  type: 'none',
+  index: 0,
+};
+
+const blurSettings = {
+  type: 'blur',
+};
+
+describe('The useBackgroundSettings hook ', () => {
```
charliesantos, line 24: Let's add test for the following use cases:

- removeProcessor
  - should call videoTrack.removeProcessor if videoTrack and videoTrack.processor exists
  - should not call videoTrack.removeProcessor if videoTrack.processor doesn't exists
  - should not call videoTrack.removeProcessor if videoTrack doesn't exists
- setBackgroundSettings
  - should call videoTrack.removeProcessor if videoTrack and videoTrack.processor exists
  - should not call videoTrack.removeProcessor if videoTrack.processor doesn't exists
  - should not call videoTrack.removeProcessor if videoTrack doesn't exists
  - should not call videoTrack.addProcessor with a param of blurProcessor if backgroundSettings.type is not equal to 'blur'
  - should not call videoTrack.addProcessor if backgroundSettings.type is 'blur' but videoTrack is missing
  - should set localStorage item for each of the use cases above

src/components/VideoProvider/useBackgroundSettings/useBackgroundSettings.test.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  it('should return the backgroundsettings and update function.', () => {
+    const { result } = renderHook(() => useBackgroundSettings(mockVideoTrack as any));
+    expect(result.current).toEqual([defaultSettings, expect.any(Function), expect.any(Function)]);
+  });
+
+  it('should set the background settings correctly and set the video processor when "blur" is selected', async () => {
+    const { result } = renderHook(() => useBackgroundSettings(mockVideoTrack as any));
+    await act(async () => {
+      result.current[1](blurSettings as BackgroundSettings);
```
charliesantos, line 43: Line 41 seems to be duplicated on all tests. I think this can be in the beforeEach hook.
Also, it's better to assign the results into variables for readability, instead of accessing them directly from the array. I think this can also live in a beforeEach hook. Something like below.

```ts
  let backgroundSettings: any;
  let setBackgroundSettings: any;
  let removeProcessor: any;
  let renderResult: any;

  beforeEach(() => {
    const { result } = renderHook(() => useBackgroundSettings(mockVideoTrack as any));
    renderResult = result;
    [backgroundSettings, setBackgroundSettings, removeProcessor] = renderResult.current;
  });
```

Then your tests would look something like

```ts
  it('should return the backgroundsettings and update function.', () => {
    expect(renderResult.current).toEqual([defaultSettings, expect.any(Function), expect.any(Function)]);
  });

  it('should set the background settings correctly and set the video processor when "blur" is selected', async () => {
    await act(async () => {
      setBackgroundSettings(blurSettings as BackgroundSettings);
    });
    expect(backgroundSettings.type).toEqual('blur');
    expect(mockVideoTrack.addProcessor).toHaveBeenCalled();
  });
```

src/components/Buttons/EndCallButton/EndCallButton.tsx
```diff
@@ -20,10 +20,15 @@ const useStyles = makeStyles((theme: Theme) =>
 
 export default function EndCallButton(props: { className?: string }) {
   const classes = useStyles();
-  const { room } = useVideoContext();
+  const { room, removeProcessor } = useVideoContext();
+
+  const handleDisconnect = (): void => {
+    removeProcessor();
```
charliesantos, line 26: I just realized the background is also applied in the prejoin screen when I refresh. I think instead of calling the remove processor here, maybe we should call it in the prejoin screen.

charliesantos (changes requested):

charliesantos (approved):
Great work Gabe! 🎉

PR #586, opened 2021-09-09
Add link to compatible browsers

jonra1993:
**Contributing to Twilio**

### JIRA link(s):

### Description

This contribution adds buttons with direct access to supported devices when the app is used on mobile devices. It can help to improve the user experience when the app is open in unsupported browsers both in Android and iOS.

WhatsApp Image 2021-09-09 at 6 22 37 PM
WhatsApp Image 2021-09-09 at 6 22 58 PM

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


timmydoza:
Hi @jonra1993 - thanks so much for the PR!

I love this idea - and the UI looks great. Unfortunately, I'm not sure that we'd be able to get this one merged in. There are a couple of points here:

- On Android, even though this app works in both Chrome and Firefox, the UI here is only giving the option to switch over to Chrome.
- On iOS, the app works in both Safari and Chrome, but the UI here only gives the user the option to switch over to Safari (but to be fair, Twilio Video support for iOS Chrome is fairly new, and it wasn't available at the time you submitted this PR).

Since the percentage of mobile users that use unsupported browsers is quite low, I'm ok to stick with our current UI where we simply direct the user to our existing [supported browsers](https://www.twilio.com/docs/video/javascript#supported-browsers) table, and leave it up to the user to choose a new browser. I'm not sure we are comfortable with a UI that seems to prefer one browser over another. I think it's important here for us to be impartial with respect to what browser our users should use. 

Thanks again for submitting the PR!

jonra1993:
Thanks, @timmydoza for the feedback I will test the last twilio version. I chose those browsers due to at the time I made this PR they were working perfectly with my users and I was forced to add these buttons in our solution due to the default browser on the phone of our users; in some cases was not compatible with WebRTC and it was easier for them just point to any compatible browser.

Issue #604, opened 2021-10-07
Landscape/Portrait issues in iOS Chrome


LucaProvencal:
**Describe the bug**
On Chrome on iOS, when screen is rotated from landscape back to portrait, display does not fill full area

**To Reproduce**
1. Using Chrome on iOS, join a Telehealth room
2. Hold device in landscape orientation
3. Rotate back to portrait orientation
4. Display of Room does not fill full area (see image)

**Expected behavior**
The display should fill the entire area, just like it does when the room is joined in portrait mode.

**Screenshots**
I've tested on 2 devices, both with Chrome 94.0.4606.52 or higher

**iPhone 11:** 
![IMG_0678](https://user-images.githubusercontent.com/25271943/136403929-bbb03f01-8cd4-4418-807c-65af8f13874a.jpg)

**iPhone 7:**
![IMG_0680](https://user-images.githubusercontent.com/25271943/136403931-88039932-f952-4ff0-a29c-c25aaeae6244.jpg)

**Environment (please complete the following information):**
 - iOS 14.7.1 (iPhone 7); iOS 14.8 (iPhone 11)
 - Chrome for iOS 94.0.4606
 - App Version: 0.6.0
 - SDK Version: 2.17.1
 - Node.js version: 15.14.0
 
 - User Agent (iPhone 7): Mozilla/5.0 (iPhone; CPU iPhone OS 14_7 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/94.0.4606.52 Mobile/15E148 Safari/604.1
 - User Agent (iPhone 11): Mozilla/5.0 (iPhone; CPU iPhone OS 14_8 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) CriOS/94.0.4606.76 Mobile/15E148 Safari/604.1

**Additional context**
I am investigating a fix, ~~might be related to MUI breakpoints~~ it is related to `useHeight` hook, but wanted to make you aware as well

timmydoza:
Hi @LucaProvencal - thanks for the detailed issue (and PR)!

I have seen this one in the mobile version of the app. Thanks for submitting the PR with a fix for this - we really appreciate it! 

Just wanted to let you know that we've got an eye on this one and we'll take a look when we find a moment. 

Thanks again.

mauro-oto:
@timmydoza any updates on this one? Thanks!

LucaProvencal:
[...]

PR #605, opened 2021-10-07
Address Landscape/Portrait UI issues

LucaProvencal:
Remove reliance on event listeners in favor of window.visualViewport.onresize. This fixes the issue in Chrome where height got stuck after rotating back to portrait: #604. Reference:  https://developer.mozilla.org/en-US/docs/Web/API/VisualViewport/onresize

Additionally, ensure view is scrolled to top when rotated. This fixes sporadic issues where `window.scrollY` had a nonzero value after rotating (without touching the screen). This seems to happen more often on older/slower devices. Scrolling to the top should fix these cases.

**Contributing to Twilio**

### Description

Moves addEventListener calls to visualViewport object, because this object has the 'resize' property they should be listening for. Offers a solution to https://github.com/twilio/twilio-video-app-react/issues/604. Please see screenshots in the issue for description. After this change I no longer saw the view getting stuck when switching back to Portrait view.
### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/hooks/useHeight/useHeight.test.tsx
+++ b/src/hooks/useHeight/useHeight.test.tsx
@@ -3,39 +3,105 @@ import { act, renderHook } from '@testing-library/react-hooks';
 import useHeight from './useHeight';
 
 describe('the useHeight hook', () => {
-  it('should return window.innerHeight', () => {
-    // @ts-ignore
-    window.innerHeight = 100;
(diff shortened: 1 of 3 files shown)


LucaProvencal:
I am actively testing this and revising the solution here.

PR #676, opened 2022-05-10
Video 8472 add dominant speaker

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-8472](https://issues.corp.twilio.com/browse/VIDEO-8472)

### Description

This PR adds the dominantSpeaker functionality for grid mode. The dominant speaker UI adds a green highlight around the participant's video when they are actively speaking. 

- When the dominantSpeaker is "on screen" (on the first page of the grid), their video will be outlined in green, and their position in the grid will not change.
- When the dominantSpeaker is "off screen" (not on the first page of the grid), their video will be outlined in green _and_ they will replace the least recent (oldest) dominantSpeaker on the first page. The least recent dominantSpeaker is moved to the end of the participant list.

gridModeDS

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/hooks/useGridParticipants/useGridParticipants.ts
```diff
@@ -0,0 +1,94 @@
...
+          /** Reorder the onscreen participants */
+          // Temporarily remove the newest dominant speaker:
+          newParticipantsArray.splice(newParticipantsArray.indexOf(newDominantSpeakerWithStartTime!), 1);
+          // Remove the least recent dominant speaker and replace them with the newest:
+          newParticipantsArray.splice(
+            newParticipantsArray.indexOf(leastRecentDominantSpeaker),
+            1,
+            newDominantSpeakerWithStartTime!
+          );
+          // Add the least recent dominant speaker back into the  array at the end:
```
timrozum, line 60: I think this is the only discrepancy between iOS and web. Currently iOS moves the oldest dominant speaker on the first page to the start of the second page. This way the entire grid is ordered by dominant speaker time + connection time which seems intuitive. I am thinking this is how zoom works.

olipyskoty: @timmydoza I think what @timrozum did with iOS makes sense. However, it is kind of tricky with the current implementation for the pagination feature. Since we are filling each page with the number of `maxGridParticipants`, the first participant on the second page may actually be an `onscreenParticipant` from the first page. 

For instance, if a room has 13 participants, and `maxGridParticipants` is 9, the first participant on the second page will actually be the fifth participant from the first page (when on iOS the first participant on the second page would be the 10th participant). Did we come to a decision about updating  the `usePagination` hook to mirror was iOS does?

src/hooks/useGridParticipants/useGridParticipants.ts
```diff
@@ -0,0 +1,94 @@
...
+
+interface OrderedParticipant {
+  participant: RemoteParticipant;
+  dominantSpeakerStartTime: number;
+}
+
+export default function useGridParticipants() {
+  const [orderedParticipants, setOrderedParticipants] = useState<OrderedParticipant[]>([]);
+  const { room } = useVideoContext();
+  const dominantSpeaker = useDominantSpeaker(true);
```
timmydoza, line 21: Do we want to `includeNull` here?

timmydoza (changes requested):
Nice work! It functions great, and the binary search stuff looks awesome. 

I've been thinking about the use of `useCollaborationParticipants` in `useMainParticipant`, `MenuBar`, and `ParticipantAudioTracks`. None of those hooks/components need to know when the domainant speaker changes, but `useCollaborationParticipants` adds a listener to that event and updates its state when the dominant speaker changes. 

I'm wondering if we could use a plain `useParticipants` hooks in these instances, where the array of participants is not updated when the dominant speaker is updated. I originally thought that a plain `useParticipants` hook could also provide an array of participants to the `useCollaborationParticipants` and `useGridParticipants` hooks, but I think that might be a bit trickier than I first thought.

So I'm going to think about this a little more. There's an opportunity here to get rid of some rerenders, but I'm not sure if it's really worth the effort. We can chat more about it later.

src/hooks/useGridParticipants/useGridParticipants.ts
olipyskoty: Nope, looks like it isn't needed. I thought it was so that when the dominant speaker stops talking (and becomes null), they wouldn't be highlighted in green any more. Now I realize that by passing `includeNull` in the `<GridView />` component, this is taken care of.

timmydoza (approved):
Looks good! My only thought is that some comments in useGridParticipant still refer to 'onscreen' participants, while the new variable name is `firstPageParticipants`.

PR #703, opened 2022-06-15
Fix memory leak in AudioLevelIndicator

timmydoza:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-10214](https://issues.corp.twilio.com/browse/VIDEO-10214)

### Description

This PR fixes an issue where the AudioLevelIndicator can cause a memory leak after it is mounted/unmounted many times. 

To see the issue, join a room, and run this code in the console which will repeatedly click the "Mute Audio" button:
```
var intervalId = setInterval(() => document.querySelector('[data-cy-audio-toggle]').click(), 10);
```

To stop this, just clear the interval id:
```
clearInterval(intervalId);
```

As the local audio track is repeatedly muted and unmuted, it will cause the AudioLevelIndicator to turn on and off. While this happens, you should see the memory footprint for the tab start to grow drastically (use Chrome's task manager to see the memory usage):

image

Also, you can see the memory leak if you take a heap snapshot and search for classes with "audio" in their name:
image

You'll see that there's a huge number of `MediaStreamsAudioSourceNode` classes.

This the fix in the PR, you will no longer see 1) the memory footprint grow, or 2) the large amount of `MediaStreamsAudioSourceNode` classes in the heap.

This memory leak was caused by the fact that all AudioLevelIndicators used to share a single AudioContext. Because of this, all MediaStreamsAudioSourceNodes that were created could not be garbage collected because the single AudioContext instance was never closed. 

This this PR, each AudioLevelIndicator now uses their own AudioContext instance, which is then closed when it is no longer needed. As a result, all resources used by the AudioLevelIndicator are now appropriately garbage collected. 

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
+++ b/src/components/AudioLevelIndicator/AudioLevelIndicator.tsx
@@ -22,7 +22,12 @@ export function initializeAnalyser(stream: MediaStream) {
 
   // Here we provide a way for the audioContext to be closed.
   // Closing the audioContext allows the unused audioSource to be garbage collected.
-  stream.addEventListener('cleanup', () => audioContext.close());
+
+  stream.addEventListener('cleanup', () => {
+    if (audioContext.state !== 'closed') {
+      audioContext.close();
+    }
+  });
 
   return analyser;
 }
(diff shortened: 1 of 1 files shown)


olipyskoty (approved):
Seems to work really well!

PR #705, opened 2022-06-16
Dogfood feedback

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-10220](https://issues.corp.twilio.com/browse/AHOYAPPS-0000)

### Description

This PR addresses some feedback received during the dogfooding session for Grid View:
- Hides Grid View settings when on mobile 
- Adds a `<ParticipantProvider />` so that the presentation participants and grid participants do not re-render when users switch between Grid and Presentation View. This helps preserve the ordering of the dominant speakers as users switch between views.
- Adds styling for landscape mode for mobile devices so that it is consistent with iOS. Note, I included the custom breakpoint from the Diagnostics app so that we are including large mobile phones that can have a similar width to a tablet when put in landscape mode. We might have to do some testing with different devices to ensure that all have the same behavior when in landscape.

dogfoodFeedback

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
+++ b/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
@@ -11,6 +11,8 @@ import {
   Button,
   Theme,
   DialogTitle,
+  useMediaQuery,
+  useTheme,
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import VideoInputList from './VideoInputList/VideoInputList';
@@ -49,6 +51,8 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 export default function DeviceSelectionDialog({ open, onClose }: { open: boolean; onClose: () => void }) {
   const classes = useStyles();
+  const theme = useTheme();
+  const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
 
   return (
     <Dialog open={open} onClose={onClose} classes={{ paper: classes.paper }}>
@@ -72,12 +76,14 @@ export default function DeviceSelectionDialog({ open, onClose }: { open: boolean
           <AudioOutputList />
         </div>
         <Divider />
-        <div className={classes.listSection}>
-          <Typography variant="h6" className={classes.headline}>
-            Grid View
-          </Typography>
-          <MaxGridParticipants />
-        </div>
+        {!isMobile && (
+          <div className={classes.listSection}>
+            <Typography variant="h6" className={classes.headline}>
+              Grid View
+            </Typography>
+            <MaxGridParticipants />
+          </div>
+        )}
       </DialogContent>
       <Divider />
       <DialogActions>
--- a/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
+++ b/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
@@ -11,8 +11,7 @@ import {
   Button,
   Theme,
   DialogTitle,
-  useMediaQuery,
-  useTheme,
+  Hidden,
 } from '@material-ui/core';
 import { makeStyles } from '@material-ui/core/styles';
 import VideoInputList from './VideoInputList/VideoInputList';
@@ -51,8 +50,6 @@ const useStyles = makeStyles((theme: Theme) => ({
 
 export default function DeviceSelectionDialog({ open, onClose }: { open: boolean; onClose: () => void }) {
   const classes = useStyles();
-  const theme = useTheme();
-  const isMobile = useMediaQuery(theme.breakpoints.down('sm'));
 
   return (
     <Dialog open={open} onClose={onClose} classes={{ paper: classes.paper }}>
@@ -76,14 +73,14 @@ export default function DeviceSelectionDialog({ open, onClose }: { open: boolean
           <AudioOutputList />
         </div>
         <Divider />
-        {!isMobile && (
+        <Hidden smDown>
           <div className={classes.listSection}>
             <Typography variant="h6" className={classes.headline}>
               Grid View
             </Typography>
             <MaxGridParticipants />
           </div>
-        )}
+        </Hidden>
       </DialogContent>
       <Divider />
       <DialogActions>
--- a/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
+++ b/src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
@@ -72,8 +72,8 @@ export default function DeviceSelectionDialog({ open, onClose }: { open: boolean
         <div className={classes.listSection}>
           <AudioOutputList />
         </div>
-        <Divider />
         <Hidden smDown>
+          <Divider />
           <div className={classes.listSection}>
             <Typography variant="h6" className={classes.headline}>
               Grid View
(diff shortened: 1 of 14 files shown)


src/components/MobileGridView/MobileGridView.tsx
```diff
@@ -1,8 +1,8 @@
 import { CSSProperties } from 'react';
-import { makeStyles, createStyles, Theme } from '@material-ui/core';
+import { makeStyles, createStyles, Theme, useTheme, useMediaQuery } from '@material-ui/core';
```
timmydoza: I hate to comment on code that's not part of a PR, but I was just thinking about the pagination indicators. What if we only adjust the layout to leave space for them only when they are present? Like this:

page-indicators

This CSS seems to do the trick:
```
const useStyles = makeStyles((theme: Theme) =>
  createStyles({
    participantContainer: {
      background: theme.gridViewBackgroundColor,
      position: 'absolute',
      top: 0,
      right: 0,
      bottom: 0,
      left: 0,
      height: '100%',
      '& .swiper': {
        height: '100%',
        '--swiper-pagination-bullet-inactive-color': 'white',
      },
      '& .swiper-wrapper': {
        height: '100%',
      },
      '& .swiper-pagination.swiper-pagination-bullets': {
        bottom: '5px',
      },
    },
    isPaginationActive: {
      '& .swiper-slide': { // To leave room for the pagination indicators
        height: 'calc(100% - 21px)', 
        paddingBottom: '21px',
      },
    },
    swiperSlide: {
      display: 'flex',
      flexWrap: 'wrap',
      alignSelf: 'center',
      alignContent: 'flex-start',
    },
  })
);
```

and then with clsx:
```
<div className={clsx(classes.participantContainer, { [classes.isPaginationActive]: remoteParticipantCount > 5 })}>
// ...
```

What do you think?

src/components/MobileGridView/MobileGridView.tsx
```diff
@@ -76,13 +80,26 @@ export function MobileGridView() {
     boxSizing: 'border-box',
   };
 
+  const landscapeParticipantVideoStyles: CSSProperties = {
+    height: remoteParticipantCount <= 3 ? '100%' : '50%',
+    // The width of each participant's video is determined by the number of participants on the grid
+    // page. Here the array indices represent a remoteParticipantCount. If the count is 4 or greater,
+    // the width will be 33.33%
```
timmydoza, line 84:
```suggestion
```
Don't think we need to duplicate this comment here

src/components/ParticipantInfo/ParticipantInfo.tsx
```diff
@@ -22,6 +22,7 @@ const borderWidth = 2;
 const useStyles = makeStyles((theme: Theme) =>
   createStyles({
     container: {
+      isolation: 'isolate',
```
timmydoza, line 25: Also in this file can you fix 
```
margin: `${GRID_VIEW_MARGIN} - borderWidth`,` in the dominantSpeaker class?
```

And I also don't think that `filter: none` is needed on the video element styling?

olipyskoty: I wonder if that `filter:none` somehow got merged from when I was playing around with blurring the video when tracks switch off 🤔

src/components/MobileGridView/MobileGridView.tsx
olipyskoty: Awesome that looks great

src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
```diff
@@ -72,12 +73,14 @@ export default function DeviceSelectionDialog({ open, onClose }: { open: boolean
           <AudioOutputList />
         </div>
         <Divider />
```
timmydoza, line 75: I think this <Divider> needs to go in the <Hidden>, so we don't end up with a second line like this:


image

timmydoza, line 75: I think this <Divider> needs to go in the <Hidden>, so we don't end up with a second line like this:


image
image

src/components/MobileGridView/MobileGridView.tsx
```diff
@@ -76,13 +84,23 @@ export function MobileGridView() {
     boxSizing: 'border-box',
   };
 
+  const landscapeParticipantVideoStyles: CSSProperties = {
+    height: remoteParticipantCount <= 3 ? '100%' : '50%',
+    width: ['100%', '50%', '33.33%', '25%', '33.33%'][Math.min(remoteParticipantCount, 4)],
+    padding: '0.2em 0.1em',
+    boxSizing: 'border-box',
+  };
```
timmydoza, line 92: It's nice how easy this was! I was afraid it would be more complicated.

timmydoza, line 92: It's nice how easy this was! I was afraid it would be more complicated.

src/components/ParticipantProvider/index.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { RemoteParticipant } from 'twilio-video';
+import useGridParticipants from '../../hooks/useGridParticipants/useGridParticipants';
+import usePresentationParticipants from '../../hooks/usePresentationParticipants/usePresentationParticipants';
+
+/*
+ *  The hooks used by the VideoProvider component are different than the hooks found in the 'hooks/' directory. The hooks
+ *  in the 'hooks/' directory can be used anywhere in a video application, and they can be used any number of times.
+ *  the hooks in the 'VideoProvider/' directory are intended to be used by the VideoProvider component only. Using these hooks
+ *  elsewhere in the application may cause problems as these hooks should not be used more than once in an application.
+ */
```
timmydoza, line 11: Not sure about this comment, since the useGridParticipants and usePresentationParticipants hooks are not in the videoProvider/ directory. 

Not sure if we need to move them here though? We could, because developers won't need to ever user them anymore (thanks to this provider), but also there's no harm in them being used elsewhere, which isn't the case with the other hooks like useRoom or useLocalTracks. 

What do you think about where these hooks should live? In any event, it might be good to update this comment to be specific to this component. We could explain why we had to move these hooks to their own provider.

timmydoza, line 11: Not sure about this comment, since the useGridParticipants and usePresentationParticipants hooks are not in the videoProvider/ directory. 

Not sure if we need to move them here though? We could, because developers won't need to ever user them anymore (thanks to this provider), but also there's no harm in them being used elsewhere, which isn't the case with the other hooks like useRoom or useLocalTracks. 

What do you think about where these hooks should live? In any event, it might be good to update this comment to be specific to this component. We could explain why we had to move these hooks to their own provider.

timmydoza (changes requested):

src/components/DeviceSelectionDialog/DeviceSelectionDialog.tsx
olipyskoty: Ah good catch!

src/components/ParticipantProvider/index.tsx
olipyskoty: I think I forgot to remove this comment when I copy/pasted the contents from the `<VideoProvider />` 🙂.

Since the hooks in the `<ParticipantProvider />` can be used anywhere in the app, I think we should probably keep them in the hooks directory. So I've gone ahead and written a new comment that explains the purpose of the `<ParticipantProvider />`

timmydoza: Ok - sounds good!

timmydoza (approved):

PR #708, opened 2022-06-22
Video 9513 update disabled video UI

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-9513](https://issues.corp.twilio.com/browse/VIDEO-9513)

### Description

This PR blurs a participant's video whenever their video is switched off. After two seconds, a short message is displayed over the blurred video letting the user know that the video has been switched off to conserve bandwidth. This is so there isn't a quick flash of the message if the video is only switched off for a brief moment. 

Also, I've updated the storybook controls so that you can unpublish/publish and switch on/switch off the video tracks individually, or a specific list of fake participants. This is the same behavior as the screenshare storybook controls.

Finally, comparing this branch against the grid view feature branch since `main` does not have storybook.

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

### DIFF
--- a/src/components/MainParticipantInfo/MainParticipantInfo.tsx
+++ b/src/components/MainParticipantInfo/MainParticipantInfo.tsx
@@ -24,6 +24,7 @@ const useStyles = makeStyles((theme: Theme) => ({
     alignItems: 'center',
     '& video': {
       filter: 'none',
+      transition: 'filter 1s cubic-bezier(0.22, 0.61, 0.36, 1)',
     },
   },
   identity: {
@@ -106,6 +107,7 @@ const useStyles = makeStyles((theme: Theme) => ({
   },
   switchedOffMessage: {
     opacity: 0,
+    textShadow: '0 0 3px rgba(0, 0, 0, 0.7)',
     animationName: '$showMessage',
     animationDuration: '0.5s',
     animationDelay: '2s',
--- a/src/components/MainParticipantInfo/MainParticipantInfo.tsx
+++ b/src/components/MainParticipantInfo/MainParticipantInfo.tsx
@@ -1,4 +1,4 @@
-import React from 'react';
+import React, { useState, useEffect } from 'react';
 import clsx from 'clsx';
 import { makeStyles, Theme } from '@material-ui/core/styles';
 import { LocalAudioTrack, LocalVideoTrack, Participant, RemoteAudioTrack, RemoteVideoTrack } from 'twilio-video';
@@ -22,6 +22,9 @@ const useStyles = makeStyles((theme: Theme) => ({
     position: 'relative',
     display: 'flex',
     alignItems: 'center',
+    '& video': {
+      filter: 'none',
+    },
   },
   identity: {
     background: 'rgba(0, 0, 0, 0.5)',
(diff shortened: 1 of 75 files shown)


src/components/MainParticipantInfo/MainParticipantInfo.tsx
```diff
@@ -104,8 +104,26 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  },
+  '@keyframes showMessage': {
+    '0%': {
+      opacity: 0,
+      visibility: 'hidden',
+    },
+    '100%': {
+      opacity: 1,
+      visibility: 'visible',
+    },
```
timmydoza, line 123: With something like this, we can give the message a nice fade in:
```
'@keyframes showMessage': {
      '0%': {
        opacity: 0,
        visibility: 'hidden',
      },
      '80%': {
        opacity: 1,
        visibility: 'visible',
      },
      '100%': {
        opacity: 1
      },
    }
```

No need to use these exact values. Play with it and see what works.

src/components/ParticipantInfo/ParticipantInfo.tsx
```diff
@@ -121,6 +134,28 @@ const useStyles = makeStyles((theme: Theme) =>
         fontSize: '0.75rem',
       },
     },
+    switchedOffMessage: {
```
timmydoza, line 137: This text could also have a shadow to make it more readable. Without it, you wouldn't be able to read white text on a white background. I don't know much about text-shadow, but something like this seemed ok: `textShadow: '0 0 3px rgba(0, 0, 0, 0.7)',`

More info here: https://css-tricks.com/almanac/properties/t/text-shadow/

timmydoza (changes requested):

src/components/MainParticipantInfo/MainParticipantInfo.tsx
olipyskoty: I like the idea of fading the message in, but I'm not sure I'm seeing much of a difference in the UI using your example. I tried changing `opacity` to something like `0.8` for `80%` and that looked kind of glitchy. I did notice that increasing the `animationDuration` from `0.5s` to `1s` created a fade in effect. I'm wondering if you're seeing something that I'm not seeing though, so you might need to show me?

timmydoza: Hmm, I wasn't originally seeing a fade in, but now I am. Not sure what changed, but it looks good now :+1:

src/components/ParticipantInfo/ParticipantInfo.tsx
```diff
@@ -136,8 +136,9 @@ const useStyles = makeStyles((theme: Theme) =>
     },
     switchedOffMessage: {
       opacity: 0,
+      textShadow: '0 0 3px rgba(0, 0, 0, 0.7)',
```
timmydoza, line 139: Same, let's add this to MainParticpantInfo too.

timmydoza (changes requested):

olipyskoty:
Created new PR in #731 to avoid large merge conflict

PR #731, opened 2022-07-13
Update track switch off UI

olipyskoty:
**Contributing to Twilio**

> All third party contributors acknowledge that any contributions they provide will be made under the same open source license that the open source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

### JIRA link(s):

- [VIDEO-9513](https://issues.corp.twilio.com/browse/VIDEO-9513)

### Description

This PR is a copy of PR #708. This branch has been created off of `main` in order to avoid a large and confusing merge conflict (merging `main` into old `VIDEO-9513-update-disabled-video-ui` branch).

### Before review
* [ ] Updated CHANGELOG.md if necessary
* [ ] Added unit tests if necessary
* [ ] Updated affected documentation
* [x] Verified locally with `npm test`
* [x] Manually sanity tested running locally
* [ ] Included screenshot as PR comment (if needed)
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary

PR #745, opened 2022-08-30
Video 10788 track switch off UI

olipyskoty:
**Contributing to Twilio**

### JIRA link(s):

- [VIDEO-10788](https://issues.corp.twilio.com/browse/VIDEO-10788)

### Description

This PR updates the track switch off UI for v2 rooms according to the new figma designs. See PR #731 for the old designs where we blurred the video.

switchoffUI

### Before review
* [x] Ready for review

### Before merge
* [ ] Got one or more +1s
* [ ] Re-tested if necessary


src/components/MainParticipantInfo/MainParticipantInfo.tsx
```diff
@@ -108,6 +108,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
   },
+  trackSwitchOffContainer: {
+    opacity: 0,
+    visibility: 'hidden',
+    transition: 'all 0.25s cubic-bezier(0.22, 0.61, 0.36, 1)',
+  },
+  isSwitchedOff: {
+    opacity: 1,
+    visibility: 'visible',
+    transition: 'all 0.5s linear 2s',
```
timmydoza, line 119: Not sure about the 2s delay. This would display a frozen video for 2 seconds before it fades to black. Seems a little long. I'm not sure I like it because it means that the UI allows the user to be confused about the frozen video for 2 seconds before we reveal what is going on. I think we might just want to let the user know right away. A fade could still be fine though.

src/components/MainParticipantInfo/MainParticipantInfo.tsx
```diff
@@ -173,7 +189,17 @@ export default function MainParticipantInfo({ participant, children }: MainParti
           </Tooltip>
         )}
       </div>
-      {(!isVideoEnabled || isVideoSwitchedOff) && (
+      <div
+        className={clsx(classes.avatarContainer, classes.trackSwitchOffContainer, {
```
timmydoza, line 193: Wondering if `avatarContainer` could be renamed to something more generic since it's now being used to contain something that isn't an avatar.

timmydoza (changes requested):
Looks good! We may want to try it out in a real room with a throttled connection.

Also, I noticed that the transitions aren't quite right when a participant is made the dominant speaker:
switch-off-transition
I'm not exactly sure what the fix here would be but we can chat about it.

src/components/MainParticipantInfo/MainParticipantInfo.tsx
```diff
@@ -108,6 +109,22 @@ const useStyles = makeStyles((theme: Theme) => ({
...
+  trackSwitchOffContainer: {
+    opacity: 0,
+    visibility: 'hidden',
+    transition: 'all 0.25s cubic-bezier(0.22, 0.61, 0.36, 1)',
+  },
+  isSwitchedOff: {
+    opacity: 1,
+    visibility: 'visible',
+    transition: 'all 0.5s linear',
+  },
```
timmydoza, line 121:
```suggestion
```
Don't think these are needed. Otherwise looks good!

timmydoza (changes requested):

PR #768, opened 2022-11-16
Add 3D Face Effects Twilio Video Processor and Menu

eludadev:
This PR adds native support for 3D face mask effects to this app. User can toggle a selection menu (like background effects) to pick from a list of face masks, and apply the chosen mask to cover their face. It uses WebGL for rendering the mask texture into the user's predicted face geometry.

The model used is TensorFlow's Face Landmarks model. The WebGL code is inspired by [Yoshiteru’s implementation](https://github.com/terryky/tfjs_webgl_app/tree/master/face_landmark), which is under an MIT license and all credits were given. Their code was modified to take advantage of TypeScript and OOP.


eludadev:
https://user-images.githubusercontent.com/101929811/202291149-f1036f96-ea00-48db-ad34-8e90f0270033.mp4

